PR #14, opened 2021-11-05
AMP-44687 Ampli dogfooding feedback updates

justin-fiedler:
WIP Node TS example only for early feedback
* Export a default ampli instance
* replaced Ampli.get/setInstance with ampli.load()
* Return promise objects from tracking methods

### DIFF
--- a/browser/typescript/react-app/public/index.html
+++ b/browser/typescript/react-app/public/index.html
@@ -25,6 +25,31 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>React App</title>
+    <script type="text/javascript">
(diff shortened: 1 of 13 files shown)


ondrej:
Really nice, summary of the things we talked about Justin (and one extra point that I don't think came up):

* Could “disabled” logic be inside core SDK
* Is there any use case for “environment” logic being in core SDK? If so maybe move it there too? Ampli would still read it but type/field would come from core SDK
* UI setting to make including API key in Ampli configurable for privacy/security
* LoadOptions could be flat if environment/disabled become part of Core SDK
* getDefaultPromiseResponse() could also be in Core SDK
* load() method isn’t using instance if it was provided by caller
* Consider deviceId under options always if “20% case”

qingzhuozhen (approved):

PR #15, opened 2021-11-16
feat: add ampli swift example

qingzhuozhen:


ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,766 @@
...
+import Amplitude
+
+public typealias MiddlewareExtra = [String: Any]
+
+public enum AmpliEnvironment: Int {
+  case development,production;
+}
+
+let ApiKey: [AmpliEnvironment: String] = [
+  .development:"00aa083ba31d20782808820370c15a71",
```
justin-fiedler, line 26: We should not commit any API keys to this this (public) repo. Can you please remove these from the source code and load from external configuration like in the TS examples (.env, .env.example)

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,766 @@
...
+  init() {
+    let propertyDict: [String: Any] = [
+      "String Const WIth Quotes": "\"String \"Const With\" Quotes\"",
+      "String Const": "String-Constant",
+      "String Int Const": 0,
+      "Integer Const": 10,
+      "Boolean Const": true,
+      "Number Const": 2.2
+    ];
+  super.init(
```
justin-fiedler, line 445: tab is off here

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,766 @@
...
+public enum AmpliEnvironment: Int {
+  case development,production;
+}
+
+let ApiKey: [AmpliEnvironment: String] = [
+  .development:"00aa083ba31d20782808820370c15a71",
+  .production:"af568af728fe7ecab9800979089ad112"
+];
+
+let DefaultOptions = [
```
justin-fiedler, line 30: Since this is only used of plan values you could make a default AMPPlan here instead of loose dictionary?

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,123 @@
...
+        //    instance.initializeApiKey("Custom api key");
+        //    ampli.load(options: LoadOptions(client: LoadClient(instance: instance)))
+        //
+        //    For testing you can disable ampli
+        //    ampli.load(options: LoadOptions(disabled: ENV.IS_TESTING ? true: false))
+        //
+        //    Make as many Ampli instances as you want
+        //    let ampli2 = new Ampli();
+        //    ampli2.load(options: LoadOptions(client: LoadClient(apiKey: "api-key-2"))
+        ampli.load(options: LoadOptions(environment: AmpliEnvironment.development, disabled: false, client: LoadClient(apiKey: ApiKey[AmpliEnvironment.development], instance: nil)));
```
justin-fiedler, line 75: Are environment and disabled required here? I see they aren't in the examples above.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,123 @@
...
+        ampli.eventNoProperties(userId: nil, extra: nil)
```
justin-fiedler, line 76: Remove userId in event methods.

justin-fiedler (changes requested):

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
qingzhuozhen: yeah, I updated to use environment inside xcode schema, let me know if this is best or we prefer other ways.

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
qingzhuozhen: Fixed.

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
qingzhuozhen: Good idea, I already replaced to create a plan object directly.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
qingzhuozhen: updated. This was created before I marked the options with default

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
qingzhuozhen: Done.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,124 @@
...
+import CoreData
+
+struct ContentView: View {
+    @Environment(\.managedObjectContext) private var viewContext
+
+    @FetchRequest(
+        sortDescriptors: [NSSortDescriptor(keyPath: \Item.timestamp, ascending: true)],
+        animation: .default)
+    private var items: FetchedResults<Item>
+    private var ampli: Ampli = Ampli()
```
justin-fiedler, line 18: I think we need a mechanism to provide a shared instance of Ampli. In JS/TS we export the default singleton instance as `ampli`, for swift I think we will need a dedicated getter e.g. `Ampli.getInstance()` or `Ampli.instance` to simplify the "common" case of a user wants to access a single instance across the entire application without having to manage the instance themselves. What do you think?

justin-fiedler (changes requested):

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
justin-fiedler: Thanks I think that works. Can you please also add a README to the sample that tell users how to set the external API keys along with `pod install` open `xcworkspace` instructions for newbies.

qingzhuozhen: Sounds good, let me add one README

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,124 @@
...
+        //
+        //    For testing you can disable ampli
+        //    ampli.load(options: LoadOptions(disabled: ENV.IS_TESTING ? true: false))
+        //
+        //    Make as many Ampli instances as you want
+        //    let ampli2 = new Ampli();
+        //    ampli2.load(options: LoadOptions(client: LoadClientOptions(apiKey: "api-key-2")))
+        let apiKey = ProcessInfo.processInfo.environment["AMPLITUDE_API_KEY"];
+        ampli.load(options: LoadOptions(client: LoadClientOptions(apiKey: apiKey)));
+        ampli.eventNoProperties(extra: nil)
```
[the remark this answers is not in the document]
qingzhuozhen: updated.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
qingzhuozhen: Added `Ampli.instance` as the default value, updated in comment. `getInstance` is more like we can support multiple single instances like how SDK does. Not sure if we want to support that though.

I left this line as it is now since the example later used load method. And one minor thing is load method now can be called multiple times and different LoadOptions can cause amplitude client to be updated differently.. Seems same for node as well. Not sure if this is really ideal though.. Maybe we should add a check on load method first?

qingzhuozhen: I just added a check after disabled, so load method can overload disabled, but not amplitude instance. Let me know your thoughts. Also support get amplitude client from ampli.

justin-fiedler: Cool, I think for now we can do a single `Ampli.instance` like you added. My hope/feeling is this is the 80+% case of users. If a user wants multiple instances they can still create them, however they need to manage the additional instances themselves. So 80% of users have a more concise interface, but power users can still get creative.

Not sure the user will want or need multiple instances of particular Ampli SDK. I think it's more likely that they would multiple tracking plans each with their own Ampli SDK wrapper instead. In this case `getInstance` wouldn't really help since there are two different Ampli classes.

Thanks for adding the initialize check in load, this is what we did in Iteratively where calling `load()` more than once would throw an error. I had intentionally omitted it here as some users had asked for the ability to call load multiple times with different options.  I agree that this is not best practice and should be avoided if possible.

I like your idea to allow resetting disabled. In the Itly SDK there was no good way to toggle-off tracking after load which made GDPR less than ideal. In this case the user can worst case access the Amplitude Client directly to disable tracking if needed.

Let's discuss more in our 1:1 today.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,124 @@
...
+                        Label("Add Item With Event Sent", systemImage: "plus")
+                    }
+                }
+            }
+            Text("Select an item")
+        }
+    }
+
+    private func addItem() {
+        //    Start by calling ampli.load()
```
justin-fiedler, line 50: Friendly reminder to clean this up after we have finalized the `Ampli.instance` usage.

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,124 @@
...
+//  ContentView.swift
+//  Shared
+//
+//  Created by Qingzhuo Zhen on 11/12/21.
+//
+
+import SwiftUI
+import CoreData
+
+struct ContentView: View {
```
justin-fiedler, line 11: Can you please add a second view that also uses Ampli to show how the instance is used across the app. e.g. I made this service in node - https://github.com/amplitude/ampli-examples/tree/main/node/typescript/src/services

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
```diff
@@ -0,0 +1,124 @@
...
+        //    Make as many Ampli instances as you want
+        //    let ampli2 = new Ampli();
+        //    ampli2.load(options: LoadOptions(client: LoadClientOptions(apiKey: "api-key-2")))
+        let apiKey = ProcessInfo.processInfo.environment["AMPLITUDE_API_KEY"];
+        ampli.load(options: LoadOptions(client: LoadClientOptions(apiKey: apiKey)));
+        ampli.eventNoProperties(extra: nil)
+        let extraDict = ["test" : "extra test"];
+        ampli.eventMaxIntForTest(properties: EventMaxIntForTestProperties(intMax10: 20), extra: extraDict);
+        ampli.eventWithConstTypes(extra: extraDict)
+        ampli.track(event: EventWithAllProperties(eventProperties: EventWithAllPropertiesProperties(optionalString: nil, requiredArray: ["array element 1", "array element 2"], requiredBoolean: true, requiredEnum: RequiredEnum.enum1, requiredInteger: 10, requiredNumber: 2.0, requiredString: "required string")), extra: nil)
```
justin-fiedler: Is extra optional here? If not can you please make it optional.

For this if you are trying to demonstrate the usage with extra then maybe set `extra: extraDict`

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,793 @@
...
+    }
+
+    self.amplitude?.setPlan(AmpliObservePlan!);
+}
+
+  public func track(event: Event, extra: MiddlewareExtra? = nil) -> Void {
+    if (!isInitializedAndEnabled()) {
+        return;
+    }
+    amplitude?.logEvent(event.eventType, withEventProperties: event.eventProperties, withMiddlewareExtra: extra as? NSMutableDictionary);
```
justin-fiedler, line 596: Is there a completion callback or promises for tracking calls in Amplitude-iOS?

justin-fiedler (changes requested):
@qingzhuozhen latest version is looking good. Hoping the remaining changes are fairly minor.

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,793 @@
...
+  }
+}
+
+public class Ampli {
+  private var amplitude: Amplitude?;
+  private(set) var disabled: Bool;
+
+  static let instance: Ampli = {
+      let defaultInstance = Ampli()
+      defaultInstance.load();
```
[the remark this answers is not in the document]
qingzhuozhen: updated

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
qingzhuozhen: Moved this to TextView now

ios/swift/AmpliSwiftSampleApp/Shared/ContentView.swift
qingzhuozhen: added a TextView and moved loading there

ios/swift/AmpliSwiftSampleApp/Shared/TextView.swift
```diff
@@ -0,0 +1,52 @@
...
+        //
+        //    For testing you can disable ampli
+        //    ampli.load(options: LoadOptions(disabled: ENV.IS_TESTING ? true: false))
+        //
+        //    Make as many Ampli instances as you want
+        //    let ampli2 = new Ampli();
+        //    ampli2.load(options: LoadOptions(client: LoadClientOptions(apiKey: "api-key-2")))
+        let apiKey = ProcessInfo.processInfo.environment["AMPLITUDE_API_KEY"];
+        let ampli = Ampli.instance
+        ampli.load(options: LoadOptions(client: LoadClientOptions(apiKey: apiKey)));
```
justin-fiedler, line 37: `Ampli.instance.load()` should be call once at app start up. Should we move it out of the TextView and into AmpliSwiftSampleAppApp? We used to call itly.load() in AppDelegate in the old Itly examples.

I am thinking we can move most of this to `AmpliSwiftSampleAppApp` and then just make one or two event calls here to show how to use the `Ampli.instance` in multiple files.

ios/swift/AmpliSwiftSampleApp/README.md
```diff
@@ -0,0 +1,33 @@
+# Ampli iOS SDK (Swift)
+An example app using the Ampli Swift SDK
+
+# Usage
+
+### Setup the project
+You will need to do the following before running the app.
+1.Set up environment variable with your API keys
+   1. Go to schema edit and find `Environment Variables` section
```
justin-fiedler, line 9:
```suggestion
   1. Go to `Edit Scheme > Run > Arguments > Environment Variables` section
```

Can you please provide a little more detail here. Ideally a person who has no idea how to use XCode would be able to get this up a running without additional information.

ios/swift/AmpliSwiftSampleApp/AmpliSwiftSampleAppTests/AmpliTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        ampli?.load(options: LoadOptions(client: LoadClientOptions(apiKey: "test-api-key")))
+    }
+    
+    func testIdenify() throws {
+        let userId = "test-user-id";
+        let deviceId = "test-device-id";
+        let identifyProperties = IdentifyProperties(optionalArray: ["optional array str"], requiredNumber: 22.0)
+        let eventOptions = EventOptions(deviceId: deviceId)
+        ampli?.amplitude?.addEventMiddleware(AMPBlockMiddleware { (payload, next) in
+            XCTAssertEqual(payload.event["event_type"] as! String, "Identify")
```
justin-fiedler, line 26: Check properties too?

ios/swift/AmpliSwiftSampleApp/AmpliSwiftSampleAppTests/AmpliTests.swift
```diff
@@ -0,0 +1,56 @@
...
+            XCTAssertEqual(payload.event["user_id"] as! String, userId)
+            XCTAssertEqual(payload.event["device_id"] as! String, deviceId)
+        })
+        ampli?.identify(userId: userId, properties: identifyProperties, options: eventOptions)
+    }
+    
+    func testTrackWithNoProperies() throws {
+        ampli?.amplitude?.addEventMiddleware(AMPBlockMiddleware { (payload, next) in
+            XCTAssertEqual(payload.event["event_type"] as! String, "Event No Properties")
```
justin-fiedler, line 35: Check props are empty?

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
```diff
@@ -0,0 +1,799 @@
...
+import Amplitude
+
+public typealias MiddlewareExtra = [String: Any]
+
+public enum AmpliEnvironment: Int {
+  case development, production;
+}
+
+let ApiKey: [AmpliEnvironment: String] = [
+  .development:"",
```
justin-fiedler, line 26: missing space

justin-fiedler (changes requested):

ios/swift/AmpliSwiftSampleApp/Shared/ampli.json
```diff
@@ -0,0 +1,8 @@
+{
```
justin-fiedler, line 1: I think we should move `ampli.json` to the root dir (same as Podfile)

ios/swift/AmpliSwiftSampleApp/Shared/TextView.swift
qingzhuozhen: @justin-fiedler Recently there are some issues reported about session related to initialize on App load inside AppDelegate. I am wondering if we should set that up inside our example, although session is not used here. Some customers wrote a good explanation about it in https://community.amplitude.com/instrumentation-and-data-management-57/ios-silent-notification-problem-how-they-mess-up-your-app-s-lifecycle-and-amplitude-session-events-492.

ios/swift/AmpliSwiftSampleApp/README.md
qingzhuozhen: Sure, let me add the suggested changes.

ios/swift/AmpliSwiftSampleApp/Shared/ampli.json
qingzhuozhen: Podfile is in root directory under app now. Is the ampli.json path set in cli?

qingzhuozhen: updated

ios/swift/AmpliSwiftSampleApp/Shared/Ampli/Ampli.swift
qingzhuozhen: fixed

ios/swift/AmpliSwiftSampleApp/AmpliSwiftSampleAppTests/AmpliTests.swift
qingzhuozhen: add a nil check

ios/swift/AmpliSwiftSampleApp/AmpliSwiftSampleAppTests/AmpliTests.swift
qingzhuozhen: updated

ios/swift/AmpliSwiftSampleApp/README.md
qingzhuozhen: updated

ios/swift/AmpliSwiftSampleApp/Shared/TextView.swift
justin-fiedler: Interesting, thanks for the article. I think to keep the example simple could still do `ampli.load()` in the App to make it clear that it should only happen once. It seems this is also what we recommend in the current Amplitude-iOS docs here - https://developers.amplitude.com/docs/ios#initialization

Hopefully anyone with session issues can look up the more advanced usage in the article you provided.

justin-fiedler (changes requested):
See ampli.load() in App comment. Otherwise LGTM.

ios/swift/AmpliSwiftSampleApp/Shared/TextView.swift
qingzhuozhen: OK. Just updated and moved to AmpliSwiftSampleAppApp for now. Just realized in Ampli we don't enable session for ios by default though.

justin-fiedler: For now let's keep the Ampli interface minimal. Users can either configure their own instance or set via `Ampli.instance.client` 

Maybe this is something we can add to LoadOptions in the future, and would be made easier if the v2 core Sdk exported a configuration interface that could be used instead of individual methods. e.g.

```
AmpConfiguration {
  serverZone: AMPServerZone,
  trackingSessionEvents: boolean;
}

Amplitude.instance().init(serverZone: AMPServerZone.EU, trackingSessionEvents: true)
```

justin-fiedler (approved):

ios/swift/AmpliSwiftSampleApp/Shared/TextView.swift
qingzhuozhen: Sounds good. A configuration interface would be more ideal for future.

PR #18, opened 2021-12-02
feat: add ios objective-c example

qingzhuozhen:


ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
```diff
@@ -0,0 +1,607 @@
...
+@property (nonatomic, strong, readwrite) NSString * _Nullable apiKey;
+@property (nonatomic, strong, readwrite) Amplitude * _Nullable instance;
++ (instancetype _Nonnull)initWithApiKey: (NSString *_Nonnull) apiKey;
++ (instancetype _Nonnull)initWithAmplitudeInstance: (Amplitude *_Nonnull)instance;
+@end
+
+@interface LoadOptions: NSObject
+@property (nonatomic, readwrite) AmpliEnvironment environment;
+@property (nonatomic, readwrite) BOOL disabled;
+@property (nonatomic, readwrite) LoadClientOptions * _Nullable client;
```
justin-fiedler, line 40: Since all of these are optional think it could make sense to switch to a builder block like we have in the Itly ObjC SDK.

https://github.com/amplitude/itly-tools/blob/master/sdk-examples/ios-objc-v2/ItlyObjCV2App/AppDelegate.m#L48-L54

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
```diff
@@ -0,0 +1,607 @@
...
++ (instancetype _Nonnull)initWithApiKey:(NSString *_Nonnull) apiKey;
++ (instancetype _Nonnull)initWithAmplitudeInstance:(Amplitude *_Nonnull)instance;
++ (instancetype _Nonnull)initWithEnvironment:(AmpliEnvironment)environment;
++ (instancetype _Nonnull)initWithDisabled:(BOOL) disabled;
++ (instancetype _Nonnull)initWithEnvironmentAndDisabled:(AmpliEnvironment)environment disabled:(BOOL) disabled;
++ (instancetype _Nonnull)initWithClient:(LoadClientOptions *_Nonnull)client;
+@end
+
+@interface EventOptions : NSObject
```
justin-fiedler, line 49: Builder block here as well. It's likely we will add all HttpV2 fields (optional) in the future which could be tedious to instantiate otherwise.

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
```diff
@@ -0,0 +1,607 @@
...
+@class EventWithOptionalPropertiesProperties;
+@class GroupProperties;
+@class IdentifyProperties;
+
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - Boxed enums
+
+/// Event 2 Property - Enum
+@interface RequiredEnum : NSObject
```
justin-fiedler, line 83: Andrey is also finding issues with Enums in Kotlin (along with some naming concerns) so I think we should create a fork of Quicktype like we talked about so we can implement these how we want. Sounds good to you?

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
```diff
@@ -0,0 +1,607 @@
...
+@interface Ampli: NSObject
+@property (nonatomic, strong, readonly) Amplitude * _Nullable amplitude;
+@property (nonatomic, assign, readwrite) BOOL disabled;
++ (instancetype _Nonnull)instance;
+- (instancetype _Nonnull )init;
+- (void)load;
+- (void)load:(LoadOptions *_Nullable)options;
+- (void)track:(Event *_Nonnull)event;
+- (void)track:(Event *_Nonnull)event withExtra:(MiddlewareExtra *_Nullable)extra;
+- (void)track:(Event *_Nonnull)event options:(EventOptions *_Nullable)options withExtra:(MiddlewareExtra *_Nullable)extra;
```
justin-fiedler, line 336: `withExtra` => `extra` to match `options` and `properties`?

ios/objective-c/AmpliObjectiveCSampleApp/Podfile
```diff
@@ -0,0 +1,21 @@
+def import_pods
+
+pod 'Amplitude', :git => 'https://github.com/amplitude/Amplitude-iOS.git', :branch => 'middleware-support'
```
justin-fiedler, line 3: Reminder to update

justin-fiedler (changes requested):

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/AppDelegate.m
```diff
@@ -0,0 +1,90 @@
...
+    [ampli eventWithConstTypes:extraDict];
+    EventWithAllPropertiesProperties *eventWithAllPropertiesProperties = [EventWithAllPropertiesProperties new];
+    eventWithAllPropertiesProperties.requiredArray = [NSArray arrayWithObjects:@"array element 1", @"array element 2", nil];
+    eventWithAllPropertiesProperties.requiredBoolean = @YES;
+    eventWithAllPropertiesProperties.requiredEnum = [RequiredEnum enum1];
+    eventWithAllPropertiesProperties.requiredInteger = @10;
+    eventWithAllPropertiesProperties.requiredNumber = @2.0;
+    eventWithAllPropertiesProperties.requiredString = @"required string";
+    EventWithAllProperties *eventWithAllProperties = [EventWithAllProperties initWithEventProperties:eventWithAllPropertiesProperties];
+    [ampli track:eventWithAllProperties];
```
justin-fiedler, line 67: This usage is pretty verbose. Could we update the event classes to use a combination of a builder block and required arguments to 1) better enforce required/optional properties 2) make the usage less verbose.

In the [Itly ObjC v2 we had this usage](https://github.com/amplitude/itly-tools/blob/master/sdk-examples/ios-objc-v2/ItlyObjCV2App/AppDelegate.m#L103-L106) which I think would be nice here as well:
```
[ampli track:[EventWithEnumTypes requiredEnum:EventWithEnumTypesRequiredEnumRequiredEnum1
                                           builderBlock:^(EventWithEnumTypesBuilder * b) {
        b.optionalEnum = EventWithEnumTypesOptionalEnumOptionalEnum1;
    }]];
```

qingzhuozhen: @justin-fiedler  One concern here is adding builderBlock for Event is not really resolving the verbose in general, as we normally encourage to use the method `eventWithAllProperties` but not the track method right? And in that case it would still be verbose as it is now. If we want to support builder pattern, I guess supporting that in properties level would be better than supporting in event level. Maybe we can make this part of the improvements we plan for quicktype.

ios/swift/AmpliSwiftSampleApp/README.md
```diff
@@ -0,0 +1,33 @@
+# Ampli iOS SDK (Swift)
```
justin-fiedler, line 1: @qingzhuozhen why are all these Swift files showing as new? Aren't they already in main?

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/AppDelegate.m
```diff
@@ -0,0 +1,90 @@
...
+    identifyProperties.optionalArray = [NSArray arrayWithObjects:@"optional string", nil];
+    identifyProperties.requiredNumber = @22;
+    [ampli identify:@"ampli-objc-user" properties:identifyProperties];
+    [ampli eventNoProperties];
+    NSMutableDictionary *extraDict = [NSMutableDictionary new];
+    [extraDict setObject:@"extra test"
+              forKey:@"test"];
+    EventMaxIntForTestProperties *eventMaxIntForTestProperties = [EventMaxIntForTestProperties new];
+    eventMaxIntForTestProperties.intMax10 = @20;
+    [ampli eventMaxIntForTest:eventMaxIntForTestProperties extra:extraDict];
```
justin-fiedler, line 57: This section is pretty dense. Can you please add some space between each Event to make it a little easier to read.

justin-fiedler (changes requested):
Looks like there are some issues with Swift. Maybe need to rebase? The swift tests are failing and all Swift files are showing as new.

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
```diff
@@ -0,0 +1,760 @@
...
+Owner: Test codegen
+
+- Parameter properties The event's properties
+*/
+- (void)eventMaxIntForTest:(EventMaxIntForTestProperties *_Nonnull)properties;
+
+/**
+Event No Properties
+
+[View in Tracking Plan](https://data.amplitude.com/test-codegen/Test%20Codegen/events/main/0.0.0/Event%20No%20Properties)
```
justin-fiedler, line 401: These links aren't clickable for me in QuickHelp even though [the format looks correct](https://stackoverflow.com/questions/43902959/hyperlink-in-an-xcode-comment). Any ideas?

image

justin-fiedler (changes requested):

ios/swift/AmpliSwiftSampleApp/README.md
qingzhuozhen: This branch is based on the earlier swift branch. The swift branch has more changes after though, will create another PR for objective-c.

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/AppDelegate.m
qingzhuozhen: Sure, will do.

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.h
qingzhuozhen: I believe the markup works for swift..However that is not same for objective-c. I have been searching and trying for hour for it..the anchor tag seems better but it is not clickable in quickhelp... Do you know if there is a way for this link in objective-c?

justin-fiedler: Unfortunately no. I also tried a bunch of things that didnt work :(

PR #25, opened 2021-12-22
feat: add group support in ios

qingzhuozhen:


ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleAppTests/AmpliTests.m
```diff
@@ -77,4 +77,19 @@ - (void)testIdentify {
     [_ampli identify:userId properties:identifyProperties options:eventOptions];
 }
 
+- (void)testSetGroup {
```
justin-fiedler, line 80: ❤️

ios/swift/AmpliSwiftSampleApp/Shared/AmpliSwiftSampleAppApp.swift
```diff
@@ -46,6 +46,9 @@ struct AmpliSwiftSampleAppApp: App {
         // Identify
         ampli.identify(userId: "ampli-swift-user", properties: IdentifyProperties(optionalArray: ["optional string"], requiredNumber: 22.0))
         
+        // Set group
+        ampli.setGroup(name: "ampli group type", value: "ampli swift group")
```
justin-fiedler, line 50: Not sure how non-Swifty it is but feel like we could make these unnamed args since they are required and the interface is unlikely to change. WDYT?

```
ampli.setGroup("ampli group type", "ampli swift group")
```

justin-fiedler (approved):

ios/swift/AmpliSwiftSampleApp/Shared/AmpliSwiftSampleAppApp.swift
qingzhuozhen: done, updated

justin-fiedler (approved):

PR #31, opened 2022-01-03
AMP-46862 Updated Kotlin Android codegen for event methods/classes

falconandy:


android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -26,442 +26,486 @@ import com.amplitude.api.Plan
...
+    fun options(options: EventOptions): Identify {
+        return Identify(this.eventType, this.eventProperties?.toMap(), options)
+    }
 }
 
-/**
- * Description for event with Array Types
- */
-data class EventWithArrayTypesProperties (
+class EventMaxIntForTest private constructor(
```
justin-fiedler, line 68: Even though it's private having the `eventType` settable in the constructor seems strange. Can we do something like the following instead?


```
class EventMaxIntForTest private constructor(
    eventProperties: Map<String, Any?>?,
    options: EventOptions? = null
) : Event(eventType = "EventMaxIntForTest", eventProperties, options) {
 ...
 constructor(
    intMax10: Int
 ) : this(mapOf(
    "intMax10" to intMax10
 ))
}
```

android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -776,7 +908,7 @@ open class Ampli {
 
             try {
                 value?.let {
-                    json.put(key, if (value is Collection<*>) JSONArray(value) else value)
+                    json.put(key, if (isArray(value)) getJsonArray(value) else value)
```
justin-fiedler, line 911: @falconandy can you please explain this change? Does `is Collection<*>` not work for certain cases?

justin-fiedler (changes requested):

android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -26,442 +26,486 @@ import com.amplitude.api.Plan
...
+            "requiredNumber" to requiredNumber
+        )
+    )
 
-    /**
-     * Event 2 Property - String
-     */
-    val requiredString: String
-)
+    fun options(userId: String? = null, deviceId: String? = null): Identify {
```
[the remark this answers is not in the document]
justin-fiedler: We could also expose the `options(userId, ...)` method on the base Event class and then it would always be in sync as both those classes will be moved to the Amplitude-Android in the long run. Not sure if we could use generics like below, if not we could just return as a generic Event.

```
class EventOptions {
    val userId: String? = null,
    val deviceId: String? = null,
}

class Event {
  fun <E:Event>options(userId: String? = null, ....): E {
    return Event(this.eventType, this.eventProperties, EventOptions(...)) as E;
  }
}
```

android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
falconandy: Fixed

android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
falconandy: `quicktype` generates `List` type - `is Collection<*>` check works. Now we returned to `Array` type and `is Collection<*>` check doesn't work

android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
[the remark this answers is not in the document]
falconandy: I did wanted to implement it in the same way but I do not like cast `Event(this.eventType, this.eventProperties, EventOptions(...)) as E` - it looks incorrect to cast base instance to child type.
Now I've fixed it in some other way.

PR #32, opened 2022-01-07
AMP-46980 Switched Ampli Swift to Itly style codegen

falconandy:
@justin-fiedler I can't test properly on Ubuntu


justin-fiedler (changes requested):
Looking good and runs without issues in XCode. See requested changes in codgen PR

PR #49, opened 2022-01-29
AMP-47972 JRE Kotlin example

falconandy:


jre/kotlin/AmpliApp/README.md
```diff
@@ -0,0 +1,30 @@
+# Ampli JRE SDK (Kotlin)
+An example app using the Ampli JRE SDK
+
+# Usage
+
+### Setup the project
+You will need to do the following before running the app.
+
+1. Create a `.env` with your API key
+    1. `cp .env.example .env`
```
justin-fiedler, line 10: Nice, using .env here is great. Can update the Android examples to match?

justin-fiedler (changes requested):
Overall looks great. See minor nit for more event tracking calls.

jre/kotlin/AmpliApp/README.md
falconandy: @justin-fiedler It seems it's possible in some way - https://github.com/cdimascio/dotenv-kotlin#android-usage. Should we use it? Maybe it's not simpler than the current `secrets.xml`

jre/kotlin/AmpliApp/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,103 @@
...
+    ampli.identify(userId, Identify(
+        requiredNumber = 42.0,
+    ))
+
+    val extra = MiddlewareExtra(mapOf("extra-key" to "extra-value"))
+    ampli.track(userId, EventWithOptionalProperties(
+        optionalBoolean = true,
+    ), extra = extra)
+
+    ampli.eventWithAllProperties(userId, EventWithAllProperties(
```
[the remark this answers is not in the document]
falconandy: Added

jre/kotlin/AmpliApp/README.md
justin-fiedler: Ok thanks, let's keep the current secrets.xml

justin-fiedler (approved):

PR #51, opened 2022-02-07
AMP-48194 Updated TS/JS codegen (added groupIdentify support/tests)

falconandy:


browser/javascript/react-app/src/ampli/index.js
```diff
@@ -331,6 +348,46 @@ export class Ampli {
...
+   * @param {MiddlewareExtra} [extra] Extra untyped parameters for use in middleware.
+   */
+  groupIdentify(groupType, groupName, properties, options, extra) {
+    if (!this.isInitializedAndEnabled()) {
+      return;
+    }
+
+    const event = {
+      event_type: SpecialEventType.Group,
+      event_properties: new Group(properties).event_properties,
```
qingzhuozhen, line 369: I don't get the purpose here. Why create a Group event, pass in properties, then read properties? I see this pattern is happening in 4 runtimes and also Identify. However, I don't get the benefit for it.

qingzhuozhen (changes requested):
Can you also add the groupIdentify call in the example?

browser/javascript/react-app/src/ampli/index.js
falconandy: @qingzhuozhen Actually it is required only if the event has some const properties. For example:
```
export class Identify implements BaseEvent {
  event_type = 'Identify';
  event_properties: IdentifyProperties & {
    'test_property_name_const_number': 123.4;
  };

  constructor(
    event_properties: IdentifyProperties,
  ) {
    this.event_properties = {
      ...event_properties,
      'test_property_name_const_number': 123.4,
    };
  }
}
```
I can add new condition to codegen templates to use simple logic if const properties do not exist.

falconandy:
@qingzhuozhen 
>  the groupIdentify call in the example

Added

node/typescript/src/ampli/index.ts
```diff
@@ -580,6 +569,31 @@ export class Ampli {
     return { promise };
   }
 
+  groupIdentify(
+    groupType: string,
+    groupName: string,
```
yuhao900914, line 574: Can node take string[] as groupName?

yuhao900914: NVM, you are right, node only allows string as gropuName.
https://github.com/amplitude/Amplitude-Node/blob/a3f6b173a6884cb11168b6d09ae3d0351533c34a/packages/identify/src/identify.ts#L54

browser/javascript/react-app/src/ampli/index.js
qingzhuozhen: That makes more sense to me now. Yeah, if we can use a simple logic to determine that would be better. If it is too hard we can also leave it.

browser/javascript/react-app/src/App.js
```diff
@@ -111,6 +111,10 @@ function App() {
           Group
         </button>
 
+        <button onClick={() => ampli.groupIdentify('test-group-2', 'a-group-value-2', { requiredBoolean: true })}>
```
qingzhuozhen, line 114: Actually, can we do groupIdentify on the same group we set with the user? Basically same group we call `setGroup`. The groupIdentify call won't automatically bind the group to the user, so I can't see the group properties set when check on the user. This would apply to other examples as well. Thank you.

qingzhuozhen (changes requested):

browser/javascript/react-app/src/ampli/index.js
falconandy: Done

browser/javascript/react-app/src/App.js
```diff
@@ -111,6 +111,10 @@ function App() {
           Group
         </button>
 
+        <button onClick={() => ampli.groupIdentify('test-group', 'a-group-value', { requiredBoolean: true })}>
```
qingzhuozhen, line 114: Hi @falconandy I guess the issue here is I still can't see the group properties with the user. Basically 
when call `setGroup('test group', 'browser-js-ampli')`, the group Identify to target the same group
`groupIdentify('test group', 'browser-js-ampli', { requiredBoolean: true })`. Or just bind the new group in setGroup call. Right now the group `a-group-value` is not shown under user profile as the user is not part of the group.

qingzhuozhen (changes requested):
Thanks for the updates on properties, that looks more cleaner now.

browser/javascript/react-app/src/App.js
qingzhuozhen: @falconandy no worries. I just made a change for the examples and it seems good now. Basically just align the group type and names between two method calls.

qingzhuozhen (approved):

PR #101, opened 2022-05-27
Amp 53227 ampli kotlin v2 example

yuhao900914:


android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+val ampli = Ampli()
+
+open class Ampli {
+    companion object {
+        val API_KEY: Map<Environment, String> = mapOf(
+            Environment.DEVELOPMENT to "",
+            Environment.PRODUCTION to ""
+        )
+
+		private val observePlan: Plan = Plan("main", "kotlin-ampli-v2", "0", "79154a50-f057-4db5-9755-775e4e9f05e6")
```
qingzhuozhen, line 428: indent

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+  }
+class EventMaxIntForTest private constructor() : BaseEvent() {
+    /**
+     * EventMaxIntForTest
+     *
+     * [View in Tracking Plan](https://data.amplitude.com/test-codegen/Test%20Codegen/events/main/latest/EventMaxIntForTest)
+     *
+     * Event to test schema validation
+     *
+     * Owner: a94a8fe5ccb19ba61c4c0873d391e987982fbbd3 06b69c130b0036944180e653185c74f600c6fe43
```
qingzhuozhen, line 97: Curious about the owner info here, are the two ids the real information?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+            }
+            else -> {
+                System.err.println("ampli.load() requires 'environment', 'client.apiKey', or 'client.instance'")
+                return
+            }
+        }
+    }
+
+    open fun track(event: BaseEvent, options: EventOptions? = null) {
+		if (!isInitializedAndEnabled()) {
```
qingzhuozhen, line 471: indent

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+     *
+     * @param intMax10 property to test schema validation
+     */
+    fun eventMaxIntForTest(
+        intMax10: Int
+    ) {
+        this.track(
+            EventMaxIntForTest(
+            intMax10 = intMax10
+        )
```
qingzhuozhen, line 531: indent

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+                return
+            }
+        }
+    }
+
+    open fun track(event: BaseEvent, options: EventOptions? = null) {
+		if (!isInitializedAndEnabled()) {
+            return
+        }
+        this._client?.track(event.eventType, this.getEventPropertiesMap(event), options)
```
qingzhuozhen, line 474: could we just call track(event, options)?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+    }
+
+    open fun identify(userId: String?, event: Identify, options: EventOptions? = null) {
+        if (!this.isInitializedAndEnabled()) {
+            return
+        }
+        this.handleIdentifyEventOptions(userId, event, options)
+        this._client?.identify(this.getEventPropertiesMap(event), options)
```
qingzhuozhen, line 482: similar as above, I wonder if we can just track directly.

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+    private fun isInitializedAndEnabled(): Boolean {
+        if (this._client == null) {
+            System.err.println("Ampli is not yet initialized. Have you called `ampli.load()` on app start?")
+            return false
+        }
+        return !this.disabled
+    }
+
+    private fun getEventPropertiesMap(event: BaseEvent?): Map<String, Any> {
+      val map = mutableMapOf<String, Any>()
```
qingzhuozhen, line 808: Seems some ident is 2 spaces and some are 4, should we align it?

android/kotlin/v2/AmpliApp/app/src/main/java/com/example/ampliapp/MainActivity.kt
```diff
@@ -0,0 +1,94 @@
...
+            ampli.track(EventWithOptionalProperties(
+                optionalBoolean = true,
+            ))
+        }
+
+        val btnEventWithAllProperties = findViewById<Button>(R.id.btn_all_properties)
+        btnEventWithAllProperties.setOnClickListener {
+            ampli.eventWithAllProperties(
+                requiredNumber = 1.23,
+                requiredArray = arrayOf("I'm", "required"),
```
qingzhuozhen, line 44: am?

android/kotlin/v2/AmpliApp/app/src/test/java/com/example/ampliapp/AmpliTest.kt
```diff
@@ -0,0 +1,198 @@
...
+    val optionsCaptor = argumentCaptor<EventOptions>()
+    val mapCaptor = argumentCaptor<Map<String, Any>>()
+    val arrayCapture = argumentCaptor<Array<String>>()
+    private val userId = "test-ampli-user-id"
+    private val deviceId = "test-ampli-device-id"
+
+    @Before
+    fun setUp() {
+        this.ampli = Ampli()
+
```
qingzhuozhen, line 30: empty line

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: I applied the local codege changes in local. For local env, instead shows the workspace, it will show ids. 
I can replace this with `Test Codegen`

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: Screen Shot 2022-05-27 at 10 21 44 PM
The indent looks right in the IDE but not right in the GitHub for some reason.

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: Same here, code in IDE looks good.

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: The function of track has 2 ways
1. `fun track(event: BaseEvent, callback: EventCallBack? = null)`
2. `fun track(eventType: String, eventProperties: Map<String, Any?>? = null, options: EventOptions? = null)`
There is no `track(event: BaseEvent, options: EventOptions? = null)`. Feels we can merge the event with options first and then call track(event) like
` open fun track(event: BaseEvent, options: EventOptions? = null) {
        if (!isInitializedAndEnabled()) {
            return
        }
        event.mergeEventOptions(options)
        this._client?.track(event)
    }`

But the options in mergeEventOptions are required to be com.amplitude.core.EventOptions no com.amplitude.android.EventOptions. So maybe we need to import com.amplitude.android.EventOptions in BaseEvent class in Kotlin? 
Screen Shot 2022-05-27 at 10 46 54 PM

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,822 @@
...
+    val client: LoadClientOptions? = null
+)
+
+class LoadClientOptions(
+    val apiKey: String? = null,
+    val instance: Amplitude? = null,
+    val plan: Plan? = null
+)
+
+class Identify private constructor() : BaseEvent() {
```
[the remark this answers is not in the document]
yuhao900914: Don't know if I understand your question right. Let me know if not. 
1. I don't think we can prevent class duplication. But when they run `ampli pull`, the auto-generated file will have conflicts, so they must deal with the duplicate name, they might change their event name.  I think for all languages, if one class is a reserved class, then we cannot re-define the class with the same name. 
2. The IdentifyEvent is not been exposed in the previous android-kotlin version. So I used the BaseEvent instead of IdentifyEvent.  Since the user may define some user properties in `Data` ui, we need to auto-generate those user properties. This class just extends the BaseEvent and overwrites the user properties. 
In Ampli we only allow one operation for user properties which is `set`.  So in order to make it works, I think we can either call an Amplitude Identify class and using `set` to set the user properties and call identify or we can set the user properties to a key value map and call the identify the call. I add a new function signature for identify call in kotlin. It's allowed to pass a map. 

In [ampli-kotlin-v1](https://github.com/amplitude/ampli-examples/blob/main/android/kotlin/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt#L63), we didn't make an identify call, where did you say we make an identify call instead?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,811 @@
...
+            return
+        }
+        val group = mutableMapOf<String, Any?>()
+        group.put(groupType, groupName)
+        event.groups = group
+        options ?. let { event.mergeEventOptions(it) }
+        this.track(event)
+    }
+
+    open fun flush() {
```
bohan-amplitude, line 512: should we also add a revenue method?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -0,0 +1,811 @@
...
+    GroupIdentify("\$groupidentify"),
+}
+
+class LoadOptions(
+    val environment: Ampli.Environment? = null,
+    val disabled: Boolean? = null,
+    val client: LoadClientOptions? = null
+)
+
+class LoadClientOptions(
```
bohan-amplitude, line 36: should we consider server zone information in tracking plan? I added a `zone` parameter in the LoadClientOptions

bohan-amplitude, line 36: should we consider server zone information in tracking plan? I added a `zone` parameter in the LoadClientOptions

android/kotlin/v2/AmpliApp/app/src/test/java/com/example/ampliapp/AmpliTest.kt
```diff
@@ -0,0 +1,197 @@
...
+            arrayOf("value-1", "value-2", "value-3"),
+            eventOptions
+        )
+
+
+        verify(client, times(1)).setGroup(eq("group-1"), arrayCapture.capture(), optionsCaptor.capture())
+        assertEquals(arrayOf("value-1", "value-2", "value-3").joinToString(), arrayCapture.allValues.first().joinToString())
+        assertEquals(userId, optionsCaptor.allValues.first().userId)
+        assertEquals(deviceId, optionsCaptor.allValues.first().deviceId)
+}
```
qingzhuozhen, line 102: nit: indent

qingzhuozhen (commented):
Maybe update the root README as well?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: @justin-fiedler , do we want to support overwriting the serverZone.

justin-fiedler: I think we should allow setting a `Configuration` rather than individual fields. Currently Ampli v1 has `LoadClientOptions.options` to set client Options we can do the same here. The naming is kinda strange given options vs configuration. Maybe we should call it `LoadClientOptions.config` or `LoadClientOptions.configuration` in v2. WDYT?

cc @bohan-amplitude @qingzhuozhen @dnshi @kevinpagtakhan

justin-fiedler (approved):
LGTM. See nit about path in README.

.github/workflows/ampli-merge-check.android.yml
```diff
@@ -19,10 +19,15 @@ jobs:
...
         uses: actions/checkout@v2
 
-      - name: (Android Kotlin) Check the Data branch is merged before merging the Git branch
-        working-directory: ./android/kotlin/AmpliApp
+      - name: (Android Kotlin V1) Check the Data branch is merged before merging the Git branch
+        working-directory: ./android/kotlin/v1/AmpliApp
         run: ampli status --is-merged -t ${{secrets.AMPLI_TOKEN}}
 
+      - name: (Android Kotlin V1) Check the Data branch is merged before merging the Git branch
+        working-directory: ./android/kotlin/v2/AmpliApp
```
justin-fiedler, line 27: This test is failing since the new runtime/seed data is not on stage. We will need to update the stage DB with the new data to fix.

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
yuhao900914: I used configuration in v2.

.github/workflows/ampli-merge-check.android.yml
yuhao900914: Separated android ampli merge check v1 and v2 into different Github actions.

justin-fiedler (changes requested):
Overall looks good. Nit on path and gradle build failing in CI.

justin-fiedler (approved):

PR #104, opened 2022-06-02
AMP-53208: Ampli python example app

bohan-amplitude:


justin-fiedler (changes requested):

python/ampli/v1/ampli/ampli.py
```diff
@@ -0,0 +1,889 @@
...
+                  event_options: Optional[EventOptions] = None):
+        if not self.initialized_and_enabled():
+            return
+        if not event_options:
+            event_options = EventOptions()
+        if user_id:
+            event_options["user_id"] = user_id
+        self._client.set_group(group_type, group_name, event_options)
+
+    def revenue(self, user_id: str,
```
qingzhuozhen, line 538: revenue? Is that required in Ampli now? cc @justin-fiedler

python/ampli/v1/ampli/ampli.py
```diff
@@ -0,0 +1,889 @@
...
+                event_options: Optional[EventOptions] = None):
+        if not self.initialized_and_enabled():
+            return
+        if not event_options:
+            event_options = EventOptions()
+        if user_id:
+            event_options["user_id"] = user_id
+        self._client.revenue(revenue_obj, event_options)
+        
+    def add(self, amp_plugin: Union[EventPlugin, DestinationPlugin]):
```
qingzhuozhen, line 549: Do we need add and remove here?

python/ampli/v1/ampli/ampli.py
```diff
@@ -0,0 +1,889 @@
...
+        """
+        event = EventNoProperties(
+
+        )
+        event['user_id'] = user_id
+        self.track(event, event_options)
+
+    def event_object_types(
+        self,
+        user_id: str,
```
qingzhuozhen, line 620: do we still need user_id if we already have event_options?

justin-fiedler: That is how we handle it in Node. Since user_id is mostly always required it was made its own param.

python/ampli/v1/ampli/ampli.py
justin-fiedler: Good call @qingzhuozhen . I think we should remove revenue here and only have load, identify, setGroup, groupIdentify, track, flush. This will reduce the potential conflicts of user defined event method names and "system" methods. They can still send revenue events with track().

README.md
```diff
@@ -26,3 +26,6 @@ See subproject README's for details on each platform.
 ## React Native
 #### [react-native/javascript](react-native/javascript/AmpliApp)
 #### [react-native/typescript](react-native/typescript/AmpliApp)
+
+## Python
+#### [python/ampli](python/ampli/v1)
```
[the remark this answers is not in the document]
bohan-amplitude: changed to python/simple/v1. Will add python/django/v1 later

README.md
```diff
@@ -26,3 +26,7 @@ See subproject README's for details on each platform.
 ## React Native
 #### [react-native/javascript](react-native/javascript/AmpliApp)
 #### [react-native/typescript](react-native/typescript/AmpliApp)
+
+## Python
+#### [python/smiple](python/simple/v1)
```
justin-fiedler, line 31: Typo `python/simple`

README.md
```diff
@@ -26,3 +26,7 @@ See subproject README's for details on each platform.
...
+#### [python/smiple](python/django/v1)
```
justin-fiedler, line 32: +1

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,818 @@
...
+}
+DEFAULT_CONFIGURATION = Config(
+    server_zone='US',
+    plan=Plan(
+        branch="main",
+        source="python-ampli",
+        version="0",
+        version_id="79154a50-f057-4db5-9755-775e4e9f05e6"
+    )
+)
```
justin-fiedler, line 110: Can you please move these to directly under the Environment class definition. I think it is helpful to have all the options and keys at the top of the file before the Events/Properties

https://github.com/amplitude/ampli-examples/blob/main/node/typescript/src/ampli/index.ts#L25-L42

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,818 @@
...
+        self.disabled: bool = False
+
+    def load(self, options: Optional[LoadOptions] = None):
+        if options:
+            self.disabled = options.disabled
+        if self.client:
+            logging.getLogger(__name__).warning('Warning: Ampli is already initialized. ampli.load() should be called once at application start up.')
+            return
+        if not options:
+            self.client = Amplitude(api_key=API_KEY[Environment.DEVELOPMENT], configuration=DEFAULT_CONFIGURATION)
```
justin-fiedler, line 442: I think we need to check if the API_KEY[Environment.DEVELOPMENT] is null, users my not have it set in Data.

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,818 @@
...
+        else:
+            env = options.environment or Environment.DEVELOPMENT
+            if options.client:
+                api_key = options.client.api_key or API_KEY[env]
+                configuration = options.client.configuration or DEFAULT_CONFIGURATION
+                self.client = options.client.instance or Amplitude(api_key=api_key, configuration=configuration)
+                self.client.configuration.api_key = api_key
+            else:
+                self.client = Amplitude(api_key=API_KEY[env], configuration=DEFAULT_CONFIGURATION)
```
justin-fiedler, line 451: Check if api key is empty here too

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,818 @@
...
+
+    def __init__(self, environment: Environment = None,
+                 disabled: bool = False,
+                 client: LoadClientOptions = None):
+        self.environment = environment
+        self.disabled = disabled
+        self.client = client
+        
+        
+class IdentifyProperties:
```
justin-fiedler, line 54: Can we have Identify/Group classes extend BaseEvent like the other Events.

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,818 @@
...
+        
+    def shutdown(self):
+        if not self.initialized_and_enabled():
+            return
+        self.client.shutdown()
+        self.disabled = True
+        
+    def event_max_int_for_test(
+        self,
+        user_id: str,
```
justin-fiedler, line 521: To allow setting userId by Event/EventOptions we should allow `user_id: str | None` (still required but can be None)

python/simple/v1/app.py
```diff
@@ -0,0 +1,63 @@
+from ampli import *
+from plugins import MyEventIDPlugin
+
+# Initialize the Ampli instance with LoadOptions and LoadClientOptions
+ampli = Ampli()
```
justin-fiedler, line 5: Can we export a default instance like we do in JS/TS? https://github.com/amplitude/ampli-examples/blob/main/node/typescript/src/ampli/index.ts#L878

python/simple/v1/app.py
```diff
@@ -0,0 +1,63 @@
...
+ampli.load(LoadOptions(Environment.DEVELOPMENT, False, client=LoadClientOptions(api_key="key", zone="EU")))
```
justin-fiedler, line 6: Can we import this key from a `.env` file or equivalent to make it easier to configure without commiting keys.
https://github.com/amplitude/ampli-examples/blob/main/node/typescript/src/app.ts#L14

python/simple/v1/app.py
```diff
@@ -0,0 +1,63 @@
...
+ampli.identify("user_id", IdentifyProperties(required_number=16.6, optional_array=['abc', 'test']))
+
+# Group Identify using GroupProperties in tracking plan
+ampli.group_identify("Org", "Engineer", GroupProperties(required_boolean=True, optional_string=None))
+
+# Set groups for user
+ampli.set_group("user_id", "Org", ["Engineer", "DevOp"])
+
+# track strongly typed event class with ampli.track
+ampli.track(EventNoProperties(), event_options=EventOptions(user_id="user_id", device_id="device_id"))
```
justin-fiedler, line 18: Missing userId as first param? Same in some other parts below.

python/simple/v1/app.py
```diff
@@ -0,0 +1,63 @@
...
+                             required_string_array=["a", "b"])
+ampli.event_with_enum_types("user_id", required_enum=EventWithEnumTypes.RequiredEnum.REQUIRED_ENUM_1,
+                            optional_enum=EventWithEnumTypes.OptionalEnum.OPTIONAL_ENUM_2)
+
+# add plugin
+my_plugin = MyEventIDPlugin()
+ampli.client.add(my_plugin)
+
+# remove plugin
+ampli.client.remove(my_plugin)
```
justin-fiedler, line 57: Could you also add a SegmentPlugin example? It is a pretty common ask.

justin-fiedler (changes requested):

python/simple/v1/ampli/ampli.py
bohan-amplitude: Core SDK expose Identify/IdentifyEvent/GroupIdentifyEvent classes. For here should I not import these classes from core SDK and implement Identify/Group class extend BaseEvent? Identify class from core SDK is used by Ampli identify/group_identify method now. Or I can add AmpliIdentifyEvent extend IdentifyEvent and AmpliGroupIdentifyEvent extend GroupIdentifyEvent?

justin-fiedler (changes requested):
Overall looks good. Can you please add PR tests like in the other ampli examples.

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,811 @@
...
+            return
+        if not event_options:
+            event_options = EventOptions()
+        if user_id:
+            event_options["user_id"] = user_id
+        event.load_event_options(event_options)
+        self.client.track(event)
+        
+    def identify(self, user_id: Optional[str], event: Identify, event_options: Optional[EventOptions] = None):
+        self.track(user_id, event, event_options)
```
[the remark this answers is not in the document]
bohan-amplitude: backend sdk don't have client level userId since it handles requests from multiple users.

PR #120, opened 2022-06-23
AMP-53208: Ampli python example app

bohan-amplitude:


python/django/v1/ampli_app/.env.example
```diff
@@ -0,0 +1,2 @@
+API_KEY=ABCD
```
justin-fiedler, line 1: Can you please name these to match the other examples https://github.com/amplitude/ampli-examples/blob/main/node/typescript/.env.example

python/simple/v1/.env.example
```diff
@@ -0,0 +1,2 @@
+API_KEY=ABCD
```
justin-fiedler, line 1: Rename to be more explicit and match other examples https://github.com/amplitude/ampli-examples/blob/main/node/typescript/.env.example

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,811 @@
...
+class Environment(enum.Enum):
+    PRODUCTION = 'production'
+    DEVELOPMENT = 'development'
+
+
+API_KEY: Dict[Environment, str] = {
+    Environment("development"): "",
+    Environment("production"): ""
+}
+DEFAULT_CONFIGURATION = Config(
```
justin-fiedler, line 39: Add a newline between these?

python/simple/v1/app.py
```diff
@@ -0,0 +1,80 @@
...
+ampli.identify("user_id", Identify(required_number=16.6, optional_array=['abc', 'test']))
+
+# Group Identify using GroupProperties in tracking plan
+ampli.group_identify("Org", "Engineer", Group(required_boolean=True, optional_string=None))
+
+# Set groups for user
+ampli.set_group("user_id", "Org", ["Engineer", "DevOp"])
+
+# track strongly typed event class with ampli.track
+ampli.track("user_id", EventNoProperties(), event_options=EventOptions(user_id="user_id", device_id="device_id"))
```
justin-fiedler, line 22: When passing user_id in EventOptions it probably makes sense to call with `ampli.track(None, EventNoProperties(), event_options=...)`

python/simple/v1/plugins.py
```diff
@@ -0,0 +1,37 @@
+from threading import RLock
+from typing import Optional
+
+from amplitude import EventPlugin, DestinationPlugin, PluginType, BaseEvent, \
+    IdentifyEvent
+from analytics import Client as SegmentClient
+
+
+class MyEventIDPlugin(EventPlugin):
```
justin-fiedler, line 9: Since these plugins can be used in all examples maybe we should move to a shared directory for easier maintaince e.g. `python/plugins/v1` ?

justin-fiedler (changes requested):

python/simple/v1/plugins.py
bohan-amplitude: I moved the plugins.py to python/plugins/v1 and tried append sys.path + importlib to import this to app.py and views.py. It's looks very ugly in apps.py and Django import external lib is harder. So I write in readme to copy the plugins module before running example apps. Is this OK?

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,812 @@
...
+    DEVELOPMENT = 'development'
+
+
+API_KEY: Dict[Environment, str] = {
+    Environment("development"): "",
+    Environment("production"): ""
+}
+
+DEFAULT_CONFIGURATION = Config(
+    server_zone='US',
```
yuhao900914, line 41: Is server_zone = US by default in core python?

python/django/v1/ampli_app/views.py
```diff
@@ -0,0 +1,110 @@
...
+from .plugins import SegmentPlugin, MyEventIDPlugin
+
+
+api_key = get_api_key('.env', 'AMPLITUDE_API_KEY')
+write_key = get_api_key('.env', 'SEGMENT_WRITE_KEY')
+ampli.load(LoadOptions(client=LoadClientOptions(api_key)))
+
+# add plugin
+my_plugin = MyEventIDPlugin()
+segment_plugin = SegmentPlugin(write_key)
```
yuhao900914, line 15: Maybe comment out the contents related to segment_plugin by default?

python/django/v1/ampli_app/util.py
```diff
@@ -0,0 +1,9 @@
+def get_api_key(env_file, key):
```
justin-fiedler, line 1: Might be better to use an existing library for this. https://pypi.org/project/python-dotenv/

python/simple/v1/README.md
```diff
@@ -0,0 +1,33 @@
...
+# Usage
+
+### Setup the project
+You will need to do the following before running the app.
+
+`pip install -r requirements.txt`
+
+`cp .env.example .env`
+
+`cp ../../plugins/v1/plugins.py plugins.py`
```
justin-fiedler, line 13: As discussed to make this easier for users lets move the plugins back into the individual examples.

I think in the future we can find a better way to share this common code across examples.

justin-fiedler (changes requested):

python/simple/v1/plugins.py
justin-fiedler: Ok, that's too bad. Let's focus on making this easy for the users to use and move the plugins back into the individual examples.

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,812 @@
...
+        version_id="79154a50-f057-4db5-9755-775e4e9f05e6"
+    )
+)
+
+
+class LoadClientOptions:
+    
+    def __init__(self, api_key: Optional[str] = None,
+                 instance: Optional[Amplitude] = None,
+                 configuration: Optional[Config] = None):
```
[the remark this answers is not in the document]
yuhao900914: Java and kotlin use configuration already. But I just checked in ampli ts and js we are still using options. 
Can we rename the configuration to options? @dnshi

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,812 @@
...
+        self.disabled: bool = False
+
+    def load(self, options: Optional[LoadOptions] = None):
+        if options:
+            self.disabled = options.disabled
+        if self.client:
+            logging.getLogger(__name__).warning('Warning: Ampli is already initialized. ampli.load() should be called once at application start up.')
+            return
+        if not options:
+            self.client = Amplitude(api_key=API_KEY[Environment.DEVELOPMENT] or API_KEY[Environment.PRODUCTION],
```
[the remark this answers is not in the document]
bohan-amplitude: re-write the load function

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,815 @@
...
+        self.disabled = options.disabled
+        if self.client:
+            logging.getLogger(__name__).warning('Warning: Ampli is already initialized. ampli.load() should be called once at application start up.')
+            return
+        if not options.client:
+            options.client = LoadClientOptions(configuration=DEFAULT_CONFIGURATION)
+        api_key = None
+        if options.client.api_key:
+            api_key = options.client.api_key
+        elif options.client.instance:
```
yuhao900914, line 450: Feels this is not needed. 
Because in line #460 if options.client.instance is not None, we do not need the api_key value.

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,815 @@
...
+        elif options.environment:
+            api_key = API_KEY[options.environment]
+        else:
+            api_key = API_KEY[Environment.DEVELOPMENT] or API_KEY[Environment.PRODUCTION]
+        if not api_key:
+            logging.getLogger(__name__).error("ampli.load() requires 'environment', 'client.api_key', or 'client.instance'")
+            return
+        configuration = options.client.configuration or DEFAULT_CONFIGURATION
+        self.client = options.client.instance or Amplitude(api_key=api_key, configuration=configuration)
+        if not self.client.configuration.plan:
```
yuhao900914, line 461: Do we need this? Should we align this on other ampli wrappers?
cc: @justin-fiedler

python/django/v1/ampli_app/ampli/ampli.py
```diff
@@ -0,0 +1,815 @@
...
+        }
+
+
+class Ampli:
+    
+    def __init__(self):
+        self.client: Amplitude = None
+        self.disabled: bool = False
+
+    def load(self, options: Optional[LoadOptions] = None):
```
justin-fiedler, line 438: Can you please add method documentation for the main methods (load, identify, group, track, etc).

python/django/v1/ampli_app/ampli/ampli.py
yuhao900914: Thanks @justin-fiedle, make sense. 
cc: @dnshi  let's also add the logic to ampli ts and js. I will add this in Kotlin.

justin-fiedler (changes requested):

python/django/v1/ampli_app/ampli/ampli.py
bohan-amplitude: updated

python/simple/v1/ampli/ampli.py
```diff
@@ -0,0 +1,863 @@
...
+
+
+class Ampli:
+    
+    def __init__(self):
+        self.client: Amplitude = None
+        self.disabled: bool = False
+
+    def load(self, options: Optional[LoadOptions] = None):
+        """Initialize the Ampli SDK. Call once when your application starts.
```
yuhao900914, line 451: Can we replace Ampli SDK with Ampli wrapper?

yuhao900914 (approved):

justin-fiedler (approved):

PR #123, opened 2022-06-30
AMP-56354 Change strong typed method to take event parameter

bohan-amplitude:


yuhao900914 (approved):

python/simple/v1/app.py
```diff
@@ -24,60 +24,90 @@
...
+        required_boolean=True,
+        required_enum=EventWithAllProperties.RequiredEnum.ENUM_2,
+        required_number=16.4,
+        required_integer=3,
+        required_string="str"
+    ))
 
 # track event with const properties
-ampli.event_with_const_types(user_id="user_id")
+ampli.event_with_const_types(user_id="user_id", event=EventWithConstTypes())
```
justin-fiedler, line 53: nit: maybe remove prop names (user_id, event) and use positional to make the example more concise. e.g. `ampli.event_with_const_types("user_id",EventWithConstTypes())`

python/simple/v1/test_ampli.py
```diff
@@ -97,7 +98,7 @@ def test_ampli_track_event_max_int_for_test_with_track_method(self):
         ampli_client = Ampli()
         ampli_client.load(LoadOptions(client=LoadClientOptions(api_key="TEST_API_KEY")))
         ampli_client.client.configuration.callback = self.callback_function
-        ampli_client.event_max_int_for_test('test_user', 9)
+        ampli_client.event_max_int_for_test('test_user', EventMaxIntForTest(9))
```
justin-fiedler, line 101: encourage named params for event props

justin-fiedler (changes requested):

python/simple/v1/app.py
falconandy: @justin-fiedler In Jre/Kotlin if an event has no properties or has only constant properties:
```
fun eventWithConstTypes(userId: String?, options: EventOptions? = null, extra: MiddlewareExtra? = null) {
    this.track(userId, EventWithConstTypes(), options, extra)
}

fun eventNoProperties(userId: String?, options: EventOptions? = null, extra: MiddlewareExtra? = null) {
    this.track(userId, EventNoProperties(), options, extra)
}
```

bohan-amplitude: I see. Should I make event with no parameter optional here too? @justin-fiedler

justin-fiedler: @bohan-amplitude Ideally event with no properties created methods without the `event` param.

fun eventWithConstTypes(userId: String?, options: EventOptions? = null) {
    this.track(userId, EventWithConstTypes(), options)
}

justin-fiedler (approved):

PR #132, opened 2022-08-17
Amp 58494 go ampli ampli example

Mercy811:


go/simple/v1/example/ampli-example.go
```diff
@@ -0,0 +1,46 @@
...
+
+func main() {
+	err := godotenv.Load()
+	if err != nil {
+		log.Fatalln("Error loading .env file")
+	}
+	apiKey := os.Getenv("AMPLITUDE_API_KEY")
+
+	// Initialize the Ampli instance with LoadOptions and LoadClientOptions
+	ampli.Client.Load(ampli.LoadOptions{
```
falconandy, line 19: I think, it looks simpler and better to use `ampli` insteaf of `ampli.Client`:
```
ampli.Load(ampli.LoadOptions{
	Client: ampli.LoadClientOptions{
		Configuration: amplitude.NewConfig(apiKey),
	},
})
defer ampli.Shutdown()

ampli.EventNoProperties("user_id", amplitude.EventOptions{})
```
If we need a new instance:
```
ampli2 := ampli.New()
ampli2.Load(ampli.LoadOptions{
	Client: ampli.LoadClientOptions{
		Configuration: amplitude.NewConfig(apiKey2),
	},
})
defer ampli2.Shutdown()

ampli2.EventNoProperties("user_id", amplitude.EventOptions{})
```

If we still want to use the singleton explicitely `ampli.Instance` looks better, I think. For example, `ampli.Instance.Client` and `ampli.Client.Client`

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+
+	"github.com/amplitude/analytics-go/amplitude"
+)
+
+var Client = Ampli{}
+
+type Environment string
+
+const (
+	DEVELOPMENT Environment = "development"
```
falconandy, line 31: Go style is `EnvironmentDevelopment` and `EnvironmentProduction`

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+	PRODUCTION  Environment = "production"
+)
+
+var APIKey = map[Environment]string{
+	DEVELOPMENT: "",
+	PRODUCTION:  "",
+}
+
+var DefaultConfiguration = amplitude.Config{
+	FlushInterval:   amplitude.DefaultFlushInterval,
```
falconandy, line 41: All members (except `Plan`) should be initialized in the core SDK

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+// Identify
+//
+// [View in Tracking Plan]: https://data.amplitude.com/test-codegen/Test%20Codegen/events/main/latest/Identify
+//
+// Identify properties
+//
+// Params:
+//	- RequiredNumber: description for Ampli.Identify RequiredNumber
+//	- OptionalArray: description for Ampli.Identify OptionalArray
+type Identify struct {
```
falconandy, line 97: In case of `Go` required and optional fields can't be distinguished. I think, event builders (like in Java) should be used instead

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+}
+
+// Shutdown disables and shutdowns Ampli Client.
+func (a *Ampli) Shutdown() {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	a.Client.Shutdown()
+	a.Disabled = true
```
falconandy, line 337: The assignment isn't thread-safe.

falconandy (changes requested):

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+			"optionalArray":  identify.OptionalArray,
+		},
+		EventOptions: eventOptions,
+	}
+
+	a.Client.Track(identifyEvent)
+}
+
+// GroupIdentify identifies a group and set group properties.
+func (a *Ampli) GroupIdentify(groupType string, groupName string, group Group, eventOptions amplitude.EventOptions) {
```
falconandy, line 286: `groupName []string`?

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+	a.Client.Track(groupIdentifyEvent)
+}
+
+// SetGroup sets group for the current user.
+func (a *Ampli) SetGroup(userID string, groupType string, groupName []string, eventOptions amplitude.EventOptions) {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	if eventOptions == (amplitude.EventOptions{}) {
```
falconandy, line 310: ???

go/simple/v1/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+		a.Client = options.Client.Instance
+	} else {
+		configuration.APIKey = apiKey
+		a.Client = amplitude.NewClient(configuration)
+	}
+}
+
+// InitializedAndEnabled checks if Ampli is initialized and enabled.
+func (a *Ampli) InitializedAndEnabled() bool {
+	return !a.Disabled
```
falconandy, line 244: `a.Client != nil`?

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
...
+	return !a.Disabled
+}
+
+// Track tracks an StronglyTypedEvent.
+func (a *Ampli) Track(userID string, event StronglyTypedEvent, eventOptions amplitude.EventOptions) {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	if userID != "" {
```
justin-fiedler, line 253: I think `nil` would be preferable to empty string `""`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+}
+
+// InitializedAndEnabled checks if Ampli is initialized and enabled.
+func (a *Ampli) InitializedAndEnabled() bool {
+	return !a.Disabled
+}
+
+// Track tracks an StronglyTypedEvent.
+func (a *Ampli) Track(userID string, event StronglyTypedEvent, eventOptions amplitude.EventOptions) {
```
justin-fiedler, line 248: We have `userId` (vs `userID`) in the other SDKs. Would be good to make it consistent.

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
...
+			"optionalArray":  identify.OptionalArray,
+		},
+		EventOptions: eventOptions,
+	}
+
+	a.Client.Track(identifyEvent)
+}
+
+// GroupIdentify identifies a group and set group properties.
+func (a *Ampli) GroupIdentify(groupType string, groupName string, group Group, eventOptions amplitude.EventOptions) {
```
justin-fiedler, line 286: GroupIdentify also requires a `userId`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
...
+		return
+	}
+
+	if eventOptions == (amplitude.EventOptions{}) {
+		return
+	}
+
+	if userID != "" {
+		eventOptions.UserID = userID
+	}
```
justin-fiedler, line 316: This block is used a lot, consider a function to make overall codegen more concise.

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
+// ampli.py
```
justin-fiedler, line 1: nit: ampli.go

justin-fiedler (changes requested):

go/simple/v1/example/ampli-example.go
justin-fiedler: +1

go/simple/v1/ampli/ampli.go
Mercy811: Deleted this block. It doesn't make any sense here

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,338 @@
...
+	DEVELOPMENT: "",
+	PRODUCTION:  "",
+}
+
+var DefaultConfiguration = amplitude.Config{
+	FlushInterval:   amplitude.DefaultFlushInterval,
+	FlushQueueSize:  amplitude.DefaultFlushQueueSize,
+	FlushMaxRetries: amplitude.DefaultFlushMaxRetries,
+	Logger:          amplitude.NewDefaultLogger(),
+	MinIDLength:     amplitude.DefaultMinIDLength,
```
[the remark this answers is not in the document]
Mercy811: Go linter recommends `ID`. All default values here are defined in core SDK. 
I can remove default prefix in the next version of core SDK.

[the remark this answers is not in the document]
Mercy811: These default values are defined in core SDK as constants. What's why they are referred with `amplitude.`

go/simple/v2/ampli/ampli.go
Mercy811: The default value of string is `""`, it returns error `mismatched types string and untyped nil` to compare a string with `nil`

go/simple/v2/ampli/ampli.go
Mercy811: It seems other SDKs don't accept `userId`. And `userId` can be passed in in `EventOptions`
- Python: https://github.com/amplitude/ampli-examples/blob/4d58124347800d4bcaa4f1045ad2e12fcbb3ec36/python/simple/v1/ampli/ampli.py#L549
- Kotlin: https://github.com/amplitude/ampli-examples/blob/4d58124347800d4bcaa4f1045ad2e12fcbb3ec36/android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt#L583
- TypeScript: https://github.com/amplitude/Amplitude-TypeScript/blob/c701a2b7ada6c8817bf28894f4d68cfa5dd0c403/packages/analytics-core/src/core-client.ts#L60

go/simple/v2/ampli/ampli.go
Mercy811: Good idea!

go/simple/v2/ampli/ampli.go
Mercy811: updated

go/simple/v1/ampli/ampli.go
Mercy811: renamed

go/simple/v1/ampli/ampli.go
Mercy811: I used all default values in core SDK. But it definitely makes it more readable to use `NewConfig` function from Core SDK here.

Updated. Use a `DefaultConfiguration()` func to return default config with default plan initialized

go/simple/v2/ampli/ampli.go
Mercy811: Can we use `ID` here? it is recommended by linter 
```
method parameter `userId` should be `userID` (golint)
```

go/simple/v1/ampli/ampli.go
Mercy811: Asked in our channel: https://amplitude.slack.com/archives/C02PMBFAZQX/p1660934701708439?thread_ts=1660932814.907109&cid=C02PMBFAZQX

`SetGroup()` takes both string and string[]. `GroupIdentify()` takes only string as group name

go/simple/v2/ampli/ampli.go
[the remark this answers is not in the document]
Mercy811: Updated to use a `DefaultConfiguration()` func to call `amplitude.NewConfig()` inside, initialize `plan`, and return the configuration

go/simple/v1/example/ampli-example.go
Mercy811: To call `ampli.Load()` directly, I can write a wrapper like this:
```
var Client = Ampli{}

type Ampli struct {
}

func (a *Ampli) Load(options LoadOptions) {
}

func Load(options LoadOptions) {
	Client.Load(options)
}
```
1. Do you think the above implementation would be a good way?
2. I should add and expose a `New()` func for initialization. Will update it

go/simple/v1/ampli/ampli.go
Mercy811: Added `NewStronglyTypedEvent()` function for every strongly typed event, which takes required properties as arguments

go/simple/v1/example/ampli-example.go
Mercy811: I was wrong. The wrapper doesn't work when it comes to `Identify()` func. We cannot declare `Identify()` func and `Identify` within the same package.

Mercy811: Renamed default Ampli to `Instance`

go/simple/v1/ampli/ampli.go
Mercy811: updated to `return !a.Disabled && a.Client != nil`

go/simple/v1/ampli/ampli.go
Mercy811: Could you give a hint why this is not thread-safe?

go/simple/v1/ampli/ampli.go
falconandy: For example, if we have an event with optional `optionalBoolean` property:
in JS/TS we have 3 options:
1. `event = {}` - `optionalBoolean` is undefined and isn't included in payload
2. `event = { optionalBooelan: false }` - `optionalBoolean` is `false` and included in payload
3. `event = { optionalBooelan: true }` - `optionalBoolean` is `true` and included in payload

With Go struct we do not have the option 1:
1. `Event{}` - `optionalBoolean` is `false` and included in payload
2. `Event{ OptionalBoolean: false }` - `optionalBoolean` is `false` and included in payload
3. `Event{ OptionalBoolean: true }` - `optionalBoolean` is `true` and included in payload

So we need something like Java builders to allow the option 1 to not include `optionalBoolean` to payload.

go/simple/v1/ampli/ampli.go
falconandy: Simplified example (`main.go`):
```
package main

import (
	"time"
)

type Ampli struct {
	Disabled bool
}

func (a *Ampli) Enabled() bool {
	return !a.Disabled
}

func (a *Ampli) Track() {
	if a.Disabled {
		return
	}
	// track
}

func (a *Ampli) Shutdown() {
	a.Disabled = false
}

func main() {
	ampli := &Ampli{}

	for i := 0; i < 10; i++ {
		go func() {
			for {
				ampli.Track()
			}
		}()
	}

	time.Sleep(time.Second)

	ampli.Shutdown()

	time.Sleep(time.Second)
}
```

Try run command:
`go run -race main.go`

Output:
```
$ go run -race main.go 
==================
WARNING: DATA RACE
Write at 0x00c00009e00f by main goroutine:
  main.(*Amplit).Shutdown()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:23 +0xe4
  main.main()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:39 +0xd9

Previous read at 0x00c00009e00f by goroutine 6:
  main.(*Amplit).Track()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:16 +0x33
  main.main.func1()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:32 +0x2e

Goroutine 6 (running) created at:
  main.main()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:30 +0x4b
==================
Found 1 data race(s)
exit status 66
falconandy@MAIN-APC:~/GolandProjects/awesomeProject1$ go run -race main.go 
==================
WARNING: DATA RACE
Write at 0x00c00011600f by main goroutine:
  main.(*Ampli).Shutdown()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:23 +0xe4
  main.main()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:39 +0xd9

Previous read at 0x00c00011600f by goroutine 7:
  main.(*Ampli).Track()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:16 +0x33
  main.main.func1()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:32 +0x2e

Goroutine 7 (running) created at:
  main.main()
      /home/falconandy/GolandProjects/awesomeProject1/main.go:30 +0x4b
==================
Found 1 data race(s)
exit status 66
```

go/simple/v2/ampli/ampli.go
falconandy: Go strings are not nullable

go/simple/v2/ampli/ampli.go
Mercy811: Wrapped this block of code within 
`func (a *Ampli) setUserID(userID string, eventOptions *amplitude.EventOptions)`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,353 @@
...
+func (a *Ampli) Identify(userID string, identify Identify, eventOptions amplitude.EventOptions) {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	a.setUserID(userID, &eventOptions)
+
+	identifyEvent := amplitude.Event{
+		EventType: amplitude.IdentifyEventType,
+		EventProperties: map[string]interface{}{
```
bohan-amplitude, line 292: Identify call should sent info in UserProperties instead of EventProperties

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,353 @@
...
+// GroupIdentify identifies a group and set group properties.
+func (a *Ampli) GroupIdentify(groupType string, groupName string, group Group, eventOptions amplitude.EventOptions) {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	groupIdentifyEvent := amplitude.Event{
+		EventType: amplitude.GroupIdentifyEventType,
+		Groups:    map[string][]string{groupType: {groupName}},
+		EventProperties: map[string]interface{}{
```
bohan-amplitude, line 311: GroupProperties instead of EventProperties

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func (stronglyTypedEvent Identify) ToEvent() amplitude.Event {
+	identify := amplitude.Identify{}
+	identify.Set("requiredNumber", stronglyTypedEvent.RequiredNumber)
+	identify.Set("optionalArray", stronglyTypedEvent.OptionalArray)
+
+	return amplitude.Event{
+		EventType:      amplitude.IdentifyEventType,
+		UserProperties: identify.Properties,
```
Mercy811, line 111: > Identify call should sent info in UserProperties instead of EventProperties
> GroupProperties instead of EventProperties

@bohan-amplitude you are right about these:
- https://github.com/amplitude/ampli-examples/pull/132#discussion_r952045432
- https://github.com/amplitude/ampli-examples/pull/132#discussion_r952045536

 Updated to use `amplitude.Identify{}.Set()` to set properties for identify event and group identify event.

go/simple/v2/ampli/ampli.go
Mercy811: updated https://github.com/amplitude/ampli-examples/pull/132#discussion_r952197111

go/simple/v2/ampli/ampli.go
Mercy811: updated https://github.com/amplitude/ampli-examples/pull/132#discussion_r952197111

go/simple/v1/ampli/ampli.go
Mercy811: Added a setter for every optional properties, for example `SetOptionalBoolean(optionalBoolean bool)`. Also added a private `optionals map[string]bool` property in struct definition to keep track of which optional properties are set. It is used in `ToEvent()` to decide whether to add the optional property to `EventProperties` or not. 

Wrote some tests in `ampli_test.go` to include the options in your comments.

go/simple/v1/ampli/ampli.go
Mercy811: Added a mutex for every occurrence of `Ampli.Disabled`. 

Tried the above example, it ended with `panic: send on closed channel`, which is caused by core sdk. Created a ticket for it: https://amplitude.atlassian.net/browse/AMP-59243

go/simple/v2/ampli/ampli.go
yuhao900914: I noticed in go we are able to directly pass userId into identify groupIdentify, setGroup, and other events. It makes sense that the identify call sets the client userId, but I don't think we should set the client userId for other track events. 
Also why do we need to set the userId for setGroup?

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+
+// GroupIdentify identifies a group and set group properties.
+func (a *Ampli) GroupIdentify(groupType string, groupName string, group *Group, eventOptions amplitude.EventOptions) {
+	group.groups = map[string][]string{groupType: {groupName}}
+
+	a.Track("", group, eventOptions)
+}
+
+// SetGroup sets group for the current user.
+func (a *Ampli) SetGroup(userID string, groupType string, groupName []string, eventOptions amplitude.EventOptions) {
```
yuhao900914, line 356: Do we need userID for the setGroup?

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+// Identify identifies a user and set user properties.
+func (a *Ampli) Identify(userID string, identify *Identify, eventOptions amplitude.EventOptions) {
+	a.Track(userID, identify, eventOptions)
+}
+
+// GroupIdentify identifies a group and set group properties.
+func (a *Ampli) GroupIdentify(groupType string, groupName string, group *Group, eventOptions amplitude.EventOptions) {
+	group.groups = map[string][]string{groupType: {groupName}}
+
+	a.Track("", group, eventOptions)
```
yuhao900914, line 352: The first input is userID, will this accidentally set the userId to an empty string?

go/simple/v2/ampli/ampli.go
Mercy811: I think server side SDKs should accept userId since different events may belong to different users. But client side SDKs can ignore userId since all events are from the same user. Previous discussion: https://amplitude.slack.com/archives/C02E1H87E2C/p1660948525608159?thread_ts=1660947850.763319&cid=C02E1H87E2C

go/simple/v2/ampli/ampli.go
Mercy811: True. Updated to use `amplitude.Track`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+
+func (stronglyTypedEvent *Identify) ToEvent() amplitude.Event {
+	identify := amplitude.Identify{}
+	identify.Set("requiredNumber", stronglyTypedEvent.RequiredNumber)
+
+	if _, present := stronglyTypedEvent.optionals["OptionalArray"]; present {
+		identify.Set("optionalArray", stronglyTypedEvent.OptionalArray)
+	}
+
+	return amplitude.Event{
```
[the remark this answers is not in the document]
Mercy811: Yes they need to accept eventOptions. 
They actually accept it here: https://github.com/amplitude/ampli-examples/pull/132#discussion_r954018819

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,385 @@
...
+
+// EventWithAllProperties
+//
+// [View in Tracking Plan]: https://data.amplitude.com/test-codegen/Test%20Codegen/events/main/latest/Event%20With%20Array%20Types
+//
+// Event with all properties
+//
+// Owner: Test codegen
+//
+//	- required_array: Event 2 Property - Array
```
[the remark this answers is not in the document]
Mercy811: Go's function comments normally don't include params. https://tip.golang.org/doc/comment
But there is no strict requirement.

Yes I forgot "Params" here. I would like to keep as consistent as other SDKs

go/simple/v2/ampli/ampli.go
Mercy811: I think for server side SDKs need to accept userId since different events may belong to different users. But for client side SDKs, it is not necessary as all events are from the same user.

More discussion: https://amplitude.slack.com/archives/C02E1H87E2C/p1660948525608159?thread_ts=1660947850.763319&cid=C02E1H87E2C

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,388 @@
...
+	group.groups = map[string][]string{groupType: {groupName}}
+	event := group.ToEvent()
+	event.EventOptions = eventOptions
+
+	a.Client.Track(event)
+}
+
+// SetGroup sets group for the current user.
+func (a *Ampli) SetGroup(userID string, groupType string, groupName []string, eventOptions amplitude.EventOptions) {
+	a.setUserID(userID, &eventOptions)
```
yuhao900914, line 360: I think it's better to move the setUserId logic into the core logic. 
cc: @justin-fiedler

Mercy811: Agree. Our current interface `Amplitude.Identify(Identify, EventOptions)` doesn't require `userid`, which is optional to pass to EventOptions.

If a user run `client.identify(identify_obj, EventOptions())` it returns an invalid event error. 

It would be better if we could update identify interface to require users to pass userid `Amplitude.Identify(userid, Identify, EventOptions)

yuhao900914: This is how we do for kotlin. 
https://github.com/amplitude/Amplitude-Kotlin/pull/37/files
For the identify call we can overwrite the eventOptions.userId = userID. 
Then in the core SDK, we can call `setUserId` if there has eventOptions.userId exist in the identify function.

go/simple/v1/ampli/ampli.go
falconandy: 1. maybe Builder pattern is more preferred to be consistent with Java codegen
2. possible alternative implementation:
```
type baseEvent struct {
	eventType  string
	properties map[string]interface{}
}

type amplitudeEvent interface {
	toAmplitudeEvent() amplitude.Event
}

func newBaseEvent(eventType string, properties map[string]interface{}) *baseEvent {
	return &baseEvent{
		eventType:  eventType,
		properties: properties,
	}
}

func (event *baseEvent) toAmplitudeEvent() amplitude.Event {
	return amplitude.Event{
		EventType:       event.eventType,
		EventProperties: event.properties,
	}
}

type Identify struct {
	*baseEvent
}

func NewIdentify(requiredNumber float64) *Identify {
	return &Identify{
		newBaseEvent(amplitude.IdentifyEventType, map[string]interface{}{
			"requiredNumber": requiredNumber,
		}),
	}
}

func (event *Identify) SetOptionalArray(optionalArray []string) *Identify {
	event.properties["optionalArray"] = optionalArray
	return event
}

func (event *Identify) toAmplitudeEvent() amplitude.Event {
	identify := amplitude.Identify{}
	for name, value := range event.properties {
		identify.Set(name, value)
	}

	return amplitude.Event{
		EventType:      event.eventType,
		UserProperties: identify.Properties,
	}
}

type Group struct {
	*baseEvent
	groups map[string][]string
}

func NewGroup(requiredBoolean bool) *Group {
	return &Group{
		baseEvent: newBaseEvent(amplitude.GroupIdentifyEventType, map[string]interface{}{
			"requiredBoolean": requiredBoolean,
		}),
	}
}

func (event *Group) SetOptionalString(optionalString string) *Group {
	event.properties["optionalString"] = optionalString
	return event
}

func (event *Group) toAmplitudeEvent() amplitude.Event {
	identify := amplitude.Identify{}
	for name, value := range event.properties {
		identify.Set(name, value)
	}

	return amplitude.Event{
		EventType:       event.eventType,
		GroupProperties: identify.Properties,
		Groups:          event.groups,
	}
}

type EventWithAllProperties struct {
	*baseEvent
}

type RequiredEnum string

const (
	RequiredEnumEnum1 RequiredEnum = "Enum1"
	RequiredEnumEnum2 RequiredEnum = "Enum2"
)

func NewEventWithAllProperties(requiredArray []string, requiredBool bool, requiredEnum RequiredEnum, requiredInteger int, requiredNumber float64, requiredString string) *EventWithAllProperties {
	return &EventWithAllProperties{
		newBaseEvent(amplitude.GroupIdentifyEventType, map[string]interface{}{
			"requiredArray":   requiredArray,
			"requiredBool":    requiredBool,
			"requiredEnum":    requiredEnum,
			"requiredInteger": requiredInteger,
			"requiredNumber":  requiredNumber,
			"requiredString":  requiredString,
		}),
	}
}

func (event *EventWithAllProperties) SetOptionalString(optionalString string) *EventWithAllProperties {
	event.properties["optionalString"] = optionalString
	return event
}

func (event *EventWithAllProperties) SetOptionalBool(optionalBool bool) *EventWithAllProperties {
	event.properties["optionalBool"] = optionalBool
	return event
}
```

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,388 @@
...
+	RequiredEnum    RequiredEnum
+	RequiredInteger int
+	RequiredNumber  float64
+	RequiredString  string
+	OptionalString  string
+	OptionalBool    bool
+	optionals       map[string]bool
+}
+
+type RequiredEnum string
```
falconandy, line 201: `EventWithAllPropertiesRequiredEnum`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,388 @@
...
+
+const (
+	RequiredEnumEnum1 RequiredEnum = "Enum1"
```
falconandy, line 204: `EventWithAllPropertiesRequiredEnumEnum1`

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,388 @@
...
+		event.EventProperties["optionalBool"] = stronglyTypedEvent.OptionalBool
+	}
+
+	return event
+}
+
+type Ampli struct {
+	Disabled bool
+	Client   amplitude.Client
+	mutex    sync.Mutex
```
falconandy, line 261: `sync.RWMutex`?

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,388 @@
...
+	a.Client.Flush()
+}
+
+// Shutdown disables and shutdowns Ampli Instance.
+func (a *Ampli) Shutdown() {
+	if !a.InitializedAndEnabled() {
+		return
+	}
+
+	a.Client.Shutdown()
```
falconandy, line 380: Should be the last line, I think. After disabling.

falconandy (changes requested):

go/simple/v2/ampli/ampli.go
falconandy: Node.js (server side):
https://github.com/amplitude/ampli-examples/blob/main/node/typescript/v2/AmpliApp/src/ampli/index.ts#L603

go/simple/v2/ampli/ampli.go
Mercy811: Make sense to add event name as prefix to avoid potential name conflict.

Updated

go/simple/v2/ampli/ampli.go
Mercy811: It prevents from potential name conflict. 
Updated.

go/simple/v2/ampli/ampli.go
Mercy811: It's better to use `RWMutex`: 
- use `RLock()` and `RUnlock()` when reading, `InitializedAndEnabled()`
- keep `Lock()`and `Unlock() when writing, `ampli.Load()` and `ampli.Shutdown()`

go/simple/v2/ampli/ampli.go
Mercy811: Correct. Current sequence may still be call track even when client is shutdown.

I should also load first, then set `Disabled` to prevent the situation where track is called but client is not ready  

Updated both

go/simple/v2/ampli/ampli.go
yuhao900914: FYI:
This setUserId is just set the eventOptions.userId. 
And since this is the server-side SDK the logic is different.

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,389 @@
...
+		a.Client = amplitude.NewClient(configuration)
+	}
+
+	a.mutex.Lock()
+	a.Disabled = options.Disabled
+	a.mutex.Unlock()
+}
+
+// InitializedAndEnabled checks if Ampli is initialized and enabled.
+func (a *Ampli) InitializedAndEnabled() bool {
```
falconandy, line 318: TypeScript version:
```
  private isInitializedAndEnabled(): boolean {
    if (!this.amplitude) {
      console.error('ERROR: Ampli is not yet initialized. Have you called ampli.load() on app start?');
      return false;
    }
    return !this.disabled;
  }
```

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	if !(apiKey != "" || options.Client.Instance != nil) {
+		log.Default().Printf("Error: Ampli.Load() requires option.Environment, " +
+			"and apiKey from either options.Instance.APIKey or APIKey[options.Environment], " +
+			"or options.Instance.Instance")
+	}
+
+	var configuration amplitude.Config
+	if !options.Client.Configuration.IsEmpty() {
```
falconandy, line 295: The logic is not simple. Did you test that all core config values (that a user do not set explicitly) will have default values (defined in core SDK)?

falconandy (changes requested):

go/simple/v1/ampli/ampli.go
Mercy811: That's awesome! 
- More consistent with Java builder pattern 
- No `optionals` in strongly typed events
- Use only one `toAmplitudeEvent()` 

Have a small question here. Why do we use pointer here
```
type EventWithAllProperties struct {
	*baseEvent
}
```
instead of 
```
type EventWithAllProperties struct {
	baseEvent
}
```
I've tried and both of them work here.

Updated

falconandy: Sorry, it was alternative implementation for your current code, not the `Builder` pattern.
The `Builder` pattern is implemented in Java code:
https://github.com/amplitude/ampli-examples/blob/main/jre/java/AmpliApp/src/main/java/com/amplitude/ampli/EventWithAllProperties.java
https://github.com/amplitude/ampli-examples/blob/main/jre/java/AmpliApp/src/main/java/org/example/App.java#L107

falconandy: `baseEvent`/`*baseEvent` doesn't matter in this case, you're right.

Mercy811: I actually quite like this implementation! Specially it uses only one `toAmplitudeEvent()`, so less code in codegen

go/simple/v2/ampli/ampli.go
Mercy811: Added tests for different load scenarios.

Only API key is required here, 5 ways to pass API key:
1. set by environment 
2. passed in `LoadClientOptions.APIKey`
3. passed in `LoadClientOptions.Client.Instance`
4. passed in `LoadClientOptions.Configuration`

`Amplitude.Client.Config.Logger` is an interface and cannot use `assert.Equal()`. To verify other default config values are set, it is equivalent to assert equal `FlushQueueSize` to default flush queue size, because all default config values are set together.

go/simple/v2/ampli/ampli.go
Mercy811: It is better to distinguish these two situations and explicitly tell users that client is not initialized.

Updated

falconandy (commented):
We separate code/events into different files only in languages that force such separation (Java for classes, Objective-C for interface/implementation files).

Mercy811:
> We separate code/events into different files only in languages that force such separation (Java for classes, Objective-C for interface/implementation files).

I put them back into one file

go/simple/v2/ampli/ampli_test.go
```diff
@@ -0,0 +1,148 @@
...
+	instance := Ampli{}
+	APIKey[EnvironmentDevelopment] = "test-development-api-key"
+	instance.Load(LoadOptions{
+		Environment: EnvironmentDevelopment,
+	})
+
+	// Amplitude.Client.Logger is an interface of which equality is determined based on memory address
+	// So here only check one of the default configurations, FlushQueueSize
+	assert.Equal(t, expectConfig.APIKey, instance.Client.GetConfig().APIKey)
+	assert.Equal(t, expectConfig.FlushQueueSize, instance.Client.GetConfig().FlushQueueSize)
```
falconandy, line 99: Compiler error: `instance.Client.GetConfig undefined (type amplitude.Client has no field or method GetConfig)`

go/simple/v2/examples/simple_example/simple_example.go
```diff
@@ -0,0 +1,102 @@
+package main
+
+import (
+	"ampli-example/ampli"
+	"github.com/amplitude/analytics-go/amplitude"
```
falconandy, line 5: Maybe we should add aliases to `ampli.go` for some function/types to use `ampli` package name only and do not import `github.com/amplitude/analytics-go/amplitude`
```
func NewConfig(apiKey string) amplitude.Config {
	return amplitude.NewConfig(apiKey)
}

type EventOptions = amplitude.EventOptions
```

go/simple/v2/ampli/ampli_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	instance := Ampli{}
+	APIKey[EnvironmentDevelopment] = "test-development-api-key"
+	instance.Load(LoadOptions{
+		Environment: EnvironmentDevelopment,
+	})
+
+	// Amplitude.Client.Logger is an interface of which equality is determined based on memory address
+	// So here only check one of the default configurations, FlushQueueSize
+	assert.Equal(t, expectConfig.APIKey, instance.Client.GetConfig().APIKey)
```
falconandy, line 98: Compiler error: `instance.Client.GetConfig undefined (type amplitude.Client has no field or method GetConfig)`

go/simple/v2/examples/simple_example/simple_example.go
```diff
@@ -0,0 +1,102 @@
...
+	ampli.Instance.Load(ampli.LoadOptions{
+		Client: ampli.LoadClientOptions{
+			Configuration: amplitude.NewConfig(apiKey),
+		},
+	})
+	defer ampli.Instance.Shutdown()
+
+	// Identify using IdentifyProperties in tracking plan
+	identifyObj := ampli.NewIdentify(16.6).SetOptionalArray([]string{"abc", "test"})
+	ampli.Instance.Identify("user_id", identifyObj, amplitude.EventOptions{})
```
falconandy, line 28: Maybe we should use `optional argument` pattern to not pass empty options explicitly:
```
func (a *Ampli) Identify(userID string, identify *Identify, eventOptions ...amplitude.EventOptions) {
	a.Track(userID, identify, eventOptions...)
}
```

```
ampli.Instance.Identify("user_id", identifyObj)
```

go/simple/v2/ampli/ampli.go
```diff
@@ -0,0 +1,689 @@
...
+	Disabled bool
+	Client   amplitude.Client
+	mutex    sync.RWMutex
+}
+
+// Load initializes the Ampli wrapper.
+// Call once when your application starts.
+func (a *Ampli) Load(options LoadOptions) {
+	if a.Client != nil {
+		log.Default().Printf("Warn: Ampli is already initialized. Ampli.load() should be called once at application start up.")
```
falconandy, line 521: 1. `log.Default().Printf` is the same as `log.Printf`, I believe
2. Without format placeholders/arguments `log.Print` should be used instead of `Printf`

falconandy, line 521: `ampli.Instance.Load()`?

falconandy (changes requested):

go/simple/v2/ampli/ampli_test.go
Mercy811: I import core Go SDK from local where I added `GetConfig()`

go/simple/v2/examples/simple_example/simple_example.go
Mercy811: But in that case, EventOptions would be an array. What if users pass multiple `EventOptions`? Do we only simply accept the first one?

go/simple/v2/examples/simple_example/simple_example.go
Mercy811: We had a previous discussion about it [here](https://github.com/amplitude/ampli-examples/pull/132#discussion_r949816823)

Could you elaborate on that alias? 
- Do you mean to change the package name of `package ampli` at the beginning of `ampli.go`? 
- What name do you think would be appropriate?

If we change the package name to `amplipackage` for example, to use `ampli.load()` directly, users still have to first `ampli := amplipackage.Instance`

go/simple/v2/ampli/ampli.go
Mercy811: Updated all to `log.Printf` and to `log.Print` where without placeholders

Mercy811: here `Ampli` is the struct defined in `ampli` package. IMO it's better to use `Ampli.load()` better, because users can possibly create their own `Ampli` instance. 

If it is confusing, I could add package name before `ampli.Ampli.Load()`.

Update `load()` to `Load()` with first letter uppercased

go/simple/v2/examples/simple_example/simple_example.go
falconandy: Yes, the first one.

go/simple/v2/ampli/ampli_test.go
falconandy: Usually Go code doesn't use `Get` prefix for function/methods names.

go/simple/v2/examples/simple_example/simple_example.go
falconandy: I mean now a user should have 2 imports with similar package names `ampli` and `amplitude`:
```
import (
	"ampli-example/ampli"
)
...
ampli.Instance.Load(ampli.LoadOptions{
	Client: ampli.LoadClientOptions{
		Configuration: amplitude.NewConfig(apiKey),
	},
})
defer ampli.Instance.Shutdown()
...
ampli.Instance.Identify("user_id", identifyObj, amplitude.EventOptions{})
```
It can be confusing.

With the aliases in generated `ampli.go`:
```
func NewConfig(apiKey string) amplitude.Config {
	return amplitude.NewConfig(apiKey)
}

type EventOptions = amplitude.EventOptions
```
only `ampli` package name can be used:
```
import (
	"ampli-example/ampli"
	"github.com/amplitude/analytics-go/amplitude"
)
...
ampli.Instance.Load(ampli.LoadOptions{
	Client: ampli.LoadClientOptions{
		Configuration: ampli.NewConfig(apiKey),
	},
})
defer ampli.Instance.Shutdown()
...
ampli.Instance.Identify("user_id", identifyObj, ampli.EventOptions{})

Comment (Mercy811) [reply go/simple/v2/examples/simple_example/simple_example.go]: It's more simple and concise to import only ampli package. 

Added aliases as following and related references 
```
// NewConfig is amplitude.NewConfig alias
var NewConfig = amplitude.NewConfig

// NewClient is amplitude.NewClient alias
var NewClient = amplitude.NewClient

// EventOptions is amplitude.EventOptions alias
type EventOptions = amplitude.EventOptions
```

File: go/simple/v2/ampli/ampli_test.go (hunk 31 above)
Comment (Mercy811) [reply go/simple/v2/ampli/ampli_test.go:98]: I referred to [this](https://go.dev/doc/effective_go#Getters) where use capitalized getter method when struct is private, and `Set` prefix is likely to be used. Here as we already have a `Config` struct, do not want to mix it up with `Client.Config()` getter function. 

Changed it to `Config()` if you don't recommend `Get` prefix here.

File: go/simple/v2/examples/simple_example/simple_example.go (hunk 32 above)
Comment (Mercy811) [reply go/simple/v2/examples/simple_example/simple_example.go]: changed to it optional as suggested above.

Comment (falconandy) [reply go/simple/v2/examples/simple_example/simple_example.go]: 1. Should be 
```
func (a *Ampli) EventNoProperties(userID string, event *EventNoProperties, eventOptions ...EventOptions) {
	a.Track(userID, event, eventOptions...)
}
```
instead of
```
func (a *Ampli) EventNoProperties(userID string, event *EventNoProperties, eventOptions ...EventOptions) {
	a.Track(userID, event, eventOptions[0])
}
```
2. `eventOptions[0]` fails if `eventOptions` is empty

File: go/simple/v2/examples/simple_example/simple_example.go [hunk 34]
```diff
@@ -0,0 +1,101 @@
...
+	eventNoProperties := ampli.NewEventNoProperties()
+	ampli.Instance.EventNoProperties("user-id", eventNoProperties)
+	ampli.Instance.Track("user-id", eventNoProperties)
+
+	requiredObjectArray := map[string]int{"k": 15}
+	eventObjectTypes := ampli.NewEventObjectTypes(map[string]string{"key": "value"}, []interface{}{requiredObjectArray})
+	ampli.Instance.EventObjectTypes("user-id", eventObjectTypes)
+	ampli.Instance.Track("user-id", eventObjectTypes)
+
+	eventWithAllProperties := ampli.NewEventWithAllProperties([]string{"abc", "test"}, true, ampli.EventWithAllPropertiesRequiredEnumEnum1, 3, 16.4, "str").SetOptionalString("optional-string")
```
justin-fiedler, line 52: I think to true enforce schema we need to use Java style builders. The current implementation does not enforce a schema changes statically 

For example the following schema change would continue to compile but would produce bad results.
MyEvent (v1) { prop1: string, prop2: string } 
MyEvent (v2) { prop2: string, zProp: string, } 

`ampli.NewMyEvent("prop1", "prop2")`
vs
`ampli.NewMyEvent.prop1("prop1").prop2("prop2")`

https://github.com/amplitude/ampli-examples/blob/main/jre/java/AmpliApp/src/main/java/org/example/App.java#L108-L115

justin-fiedler (changes requested):

go/simple/v2/examples/data_race_example/data_race_example.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"ampli-example/ampli"
+	"github.com/joho/godotenv"
+	"log"
+	"os"
+	"time"
```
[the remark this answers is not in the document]
falconandy: Fixed

go/simple/v2/examples/simple_example/simple_example.go
```diff
@@ -0,0 +1,101 @@
...
+	ampli.Instance.Load(ampli.LoadOptions{
+		Client: ampli.LoadClientOptions{
+			Configuration: ampli.NewConfig(apiKey),
+		},
+	})
+	defer ampli.Instance.Shutdown()
+
+	// Identify using IdentifyProperties in tracking plan
+	identifyObj := ampli.NewIdentify(16.6).SetOptionalArray([]string{"abc", "test"})
+	ampli.Instance.Identify("user_id", identifyObj)
```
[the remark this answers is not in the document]
falconandy: Fixed, but builder creation is longer

go/simple/v2/examples/simple_example/simple_example.go
falconandy: @justin-fiedler 
1. Added `builders` for 3 events - `EventNoProperties`, `EventWithAllProperties` and `EventWithOptionalProperties`
2. Event enums is now use style like `ampli.EventWithAllProperties.RequiredEnum.Enum1` instead of `ampli.EventWithAllPropertiesRequiredEnumEnum1`
3. To not pollute `ampli` namespace with intermediate builder types anonymous types are used in codegen:
```
	Builder func() interface {
		SetRequiredArray(requiredArray []string) interface {
			SetRequiredBool(requiredBool bool) interface {
				SetRequiredEnum(requiredEnum EventWithAllPropertiesRequiredEnum) interface {
					SetRequiredInteger(requiredInteger int) interface {
						SetRequiredNumber(requiredNumber float64) interface {
							SetRequiredString(requiredString string) EventWithAllPropertiesBuilder
						}
					}
				}
			}
		}
	}
```

go/simple/v2/examples/simple_example/simple_example.go
```diff
@@ -0,0 +1,113 @@
...
+	ampli.Instance.Identify(userID, ampli.Identify.Builder().
+		SetRequiredNumber(16.6).
+		SetOptionalArray([]string{"abc", "test"}).
+		Build())
+
+	// Group Identify using GroupProperties in tracking plan
+	ampli.Instance.GroupIdentify("Org", "Engineer", ampli.Group.Builder().
+		SetRequiredBoolean(true).
+		SetOptionalString("optional-string").
+		Build())
```
justin-fiedler, line 39: Thank you @falconandy the new builders are great.

nit: can we remove the `Set` prefix similar to Java builders? E.g. 
```suggestion
	ampli.Instance.GroupIdentify("Org", "Engineer", ampli.Group.Builder().
		RequiredBoolean(true).
		OptionalString("optional-string").
		Build())
```

go/simple/v2/tests/ampli_test.go
```diff
@@ -0,0 +1,108 @@
+package tests
```
justin-fiedler, line 1: Can you please add to CI workflow

justin-fiedler (changes requested):

go/simple/v2/tests/ampli_test.go
falconandy: Added

falconandy (approved):

justin-fiedler (approved):

PR #140, opened 2022-09-13
kotlin example for adding ingestion metadata plugin

yuhao900914:


android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
```diff
@@ -0,0 +1,49 @@
...
+    override fun execute(event: BaseEvent): BaseEvent? {
+
+        var libraryArr = event.library?.split("/")
+        var libraryVersion = libraryArr?.get(1)
+        println(libraryVersion)
+        println("####################")
+        if (!libraryVersion.isNullOrEmpty() && compare(libraryVersion, minIngestionMetadataSupportedVersion) >= 0) {
+            println("should set the ingestionMetadata")
+            println(libraryVersion)
+            event.ingestionMetadata
```
justin-fiedler, line 42: Will this give a compilation error in older SDKs? We need reflection or another way to dynamically access this property.

qingzhuozhen: Similar discussion with @liuyang1520 on this earlier. Maybe try on the earlier versions and verify it is not throwing compilation error.

yuhao900914: Doing the investigation about the dynamic access.

justin-fiedler: We can instead set `event.extra.amplitude = IngestionMetadata()` and then read it in Amplitude Destination Plugin.
https://github.com/amplitude/Amplitude-Kotlin/blob/main/core/src/main/java/com/amplitude/core/events/EventOptions.kt#L42

justin-fiedler: Or we could do the JSON in core:

Plugin
`event.extra["amplitude"] = { ingestionMetadata: IngestionMetadata() }`

Core:
`val ingestiontionJson = JsonObject(event.extra["amplitude"]["ingestionMetadata"])`

justin-fiedler: This approach may be better since it doesn't require reflection (slow). WDYT? @yuhao900914 @qingzhuozhen @liuyang1520

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+        if (event.extra == null) event.extra = mapOf()
+        val ingestionMetadataObject =
+            IngestionMetadata("android-kotlin-ampli", "1.0.0").toJsonObject()
+        val extra = mapOf<String, Any>().plus(
+            mapOf(
+                "amplitude" to
+                        mapOf(
+                            "ingestionMetadata" to ingestionMetadataObject
```
liuyang1520, line 18: Do we need this class (`IngestionMetadata`) if we anyways need to covert it to json object? Think it might be too much to use the class -> object -> json object.

```
JSONObject("""{"source_name":"android-kotlin-ampli", "source_version":"1.0.0"}""")
```
or
```
val jsonObject = JSONObject()
rootObject.put("source_name","android-kotlin-ampli")
rootObject.put("source_version","1.0.0")
```

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadata.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.json.JSONException
+import org.json.JSONObject
+
+class IngestionMetadata (sourceName: String, sourceVersion: String) {
+     var sourceName : String
+     var sourceVersion : String
+    init {
+        this.sourceName = sourceName
+        this.sourceVersion = sourceVersion
+    }
```
justin-fiedler, line 13: I think this is equivalent to:
```suggestion
class IngestionMetadata (var sourceName: String, var sourceVersion: String) {
```

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import com.amplitude.core.events.BaseEvent
+import com.amplitude.core.platform.Plugin
+
+class IngestionMetadataErichmentPlugin : Plugin {
+    override val type: Plugin.Type = Plugin.Type.Enrichment
+    override lateinit var amplitude: com.amplitude.core.Amplitude
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+        if (event.extra == null) event.extra = mapOf()
```
justin-fiedler, line 11: nit: single line if's can lead to trouble please add braces

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadata.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun toJsonObject(): JSONObject {
+        var jsonObject = JSONObject()
+        try {
+            if (!sourceName.isNullOrEmpty()) {
```
justin-fiedler, line 18: Is this necessary with Koltins null safety? https://kotlinlang.org/docs/null-safety.html#nullable-types-and-non-null-types

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
justin-fiedler: Can probably make the following logic more concise with this

```
val amplitudeExtra = event.extra ? event.extra["amplitude"] ?: mapOf() : mapOf();

amplitudeExta.plus(this.ingestionMetadata.toJson())

event.extra["amplitude"] = amplitudeExtra;
```

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
```diff
@@ -0,0 +1,25 @@
...
+        val ingestionMetadataObject =
+            IngestionMetadata("android-kotlin-ampli", "1.0.0").toJsonObject()
```
justin-fiedler, line 13: This should be configured externally by Ampli, not hard coded.

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -508,6 +513,7 @@ open class Ampli {
         if (this.client?.configuration?.plan == null) {
             this.client?.configuration?.plan = defaultObservePlan
         }
+        this.client?.add(IngestionMetadataErichmentPlugin())
```
justin-fiedler, line 516: Ampli should set in the sourceName and version to the Plugin.
```suggestion
        this.client?.add(IngestionMetadataErichmentPlugin(IngestionMetaData("android-kotlin-ampli", "1.0.0")))
```

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
```diff
@@ -0,0 +1,25 @@
+package com.amplitude.ampli
+
+import com.amplitude.core.events.BaseEvent
+import com.amplitude.core.platform.Plugin
+
+class IngestionMetadataErichmentPlugin : Plugin {
```
justin-fiedler, line 6: nit: spelling. Can also probably just shorten to `IngestionMetadataPlugin`

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
justin-fiedler: I think we should use `IngestionMetadata` type. I wanted it to make setting the necessary explicit, and allow for better decoupling between Ampli and Amplitude SDK.

We should also probably add accessors for post-construction configuration. plugin.setIngestionMetadata(IngestionMetadata(...))

justin-fiedler (changes requested):

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
yuhao900914: Is `event.extra["amplitude"]` a right syntax for kotlin?
Also since we didn't have any type defined fo event.extra.amplitude.  We need it to cover to the right format in order to do amplitudeExtra.plus(xxx)

Also, amplitude might have different types of info later. do we need save `ingestionMetadata` in amplitude.ingestionMetadata key?

Hope this is better. 

```
        val amplitudeExtra = (event.extra?.get("amplitude") ?: mapOf<String, Any>()) as Map<String, Any>
        val ingestionMetadataMap = mapOf(
            "ingestionMetadata" to  ingestionMetadata.toJsonObject()
        )
        amplitudeExtra.plus(ingestionMetadataMap)
        event.extra = event.extra?.plus(mapOf("amplitude" to amplitudeExtra))
        return event
```

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataPlugin.kt
```diff
@@ -0,0 +1,20 @@
...
+
+class IngestionMetadataPlugin(ingestionMetadata: IngestionMetadata) : Plugin {
+    override val type: Plugin.Type = Plugin.Type.Enrichment
+    override lateinit var amplitude: com.amplitude.core.Amplitude
+    val ingestionMetadata = ingestionMetadata
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+        val amplitudeExtra = (event.extra?.get("amplitude") ?: mapOf<String, Any>()) as Map<String, Any>
+        val ingestionMetadataMap = mapOf(
+            "ingestionMetadata" to  ingestionMetadata.toJsonObject()
```
qingzhuozhen, line 14: Do we really need this to be a JSONObject? Could it be just a map?

yuhao900914: we already has jsonObject <-> ingestionMetadata converter in core sdk. 
If we use map instead, do we also need to have a converter to convert the map to ingestionMetadata?

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/IngestionMetadataErichmentPlugin.kt
justin-fiedler: Thanks, that was pseudo code above. Didn't have the fortitude to type out the full kotlin.

PR #141, opened 2022-09-14
feat: update ingestion metadata python example

liuyang1520:
## Notes
This change adds the ingestion metadata example for Python Ampli.

## Updates
- feat: update ingestion metadata python example

## Tests
- Local echo server test, with v1.0.0 and v1.1.0

### DIFF
--- a/python/simple/v1/ampli/ampli.py
+++ b/python/simple/v1/ampli/ampli.py
@@ -49,12 +49,12 @@ DEFAULT_CONFIGURATION = Config(
 
 class LoadClientOptions:
     """Client options setting to initialize Ampli client
-    
-    :param api_key: The API key of of Amplitude project. Default to None.
+
+    :param api_key: The API key of Amplitude project. Default to None.
     :param instance: The core SDK instance used by Ampli client. Default to None
     :param configuration: The core SDK client configuration instance. Default to None
     """
-    
+
     def __init__(self, api_key: Optional[str] = None,
                  instance: Optional[Amplitude] = None,
                  configuration: Optional[Config] = None):
@@ -65,12 +65,12 @@ class LoadClientOptions:
 
 class LoadOptions:
     """Options setting to initialize Ampli client
-    
+
     :param environment: The environment of Amplitude Data project. Default to None.
     :param disabled: The flag of disabled Ampli client. Default to False
     :param client: The LoadClientOptions instance. Default to None
     """
-    
+
     def __init__(self, environment: Environment = None,
                  disabled: bool = False,
                  client: LoadClientOptions = None):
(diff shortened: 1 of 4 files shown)


python/simple/v1/ampli/ampli.py
```diff
@@ -508,7 +508,17 @@ def load(self, options: Optional[LoadOptions] = None):
         
         if not self.client.configuration.plan:
             self.client.configuration.plan = DEFAULT_CONFIGURATION.plan
-    
+
+        # Python SDK min version 1.1.0 supports IngestionMetadata
+        # This lazy module loading is for backward compatible concern.
+        import amplitude
+        if hasattr(amplitude, "IngestionMetadata"):
```
yuhao900914, line 515: This makes sense to me. It's easier than the plugin ways.

bohan-amplitude: Can you re format this like 
```
try:
    from amplitude import IngestionMetadata
except ImportError:
    IngestionMetadata = None

...
if IngestionMetadata:
    ....
```

bohan-amplitude: The import statement could go to top of the file

liuyang1520: I try to put the import in the function, to scope it, not pollute the global namespace. If we want to use `try`, think this might be better:
```python
try:
    from amplitude import IngestionMetadata
    self.client.configuration.ingestion_metadata = IngestionMetadata(
        source_name="python-python-ampli",
        source_version="1.0.0"
    )
except ImportError:
    pass
```
I don't have a preference on using `getattr(amplitude, "IngestionMetadata")` or `from amplitude import IngestionMetadata`, or even `__import__` or `importlib.import_module( )`. Let me know if you have any preference and reason, I can try that approach. Thanks!

justin-fiedler: Nice @liuyang1520 I like it

justin-fiedler (approved):
Looks good but please make sure server codegen is merged before updating the public examples.

python/simple/v1/ampli/ampli.py
bohan-amplitude: If we don't want to expose the IngestionMetadata class just use it in this method. I'm ok with either way.

python/simple/v1/ampli/requirements.txt
```diff
@@ -23,4 +23,4 @@
 # (not inside this directory)
 # Add '-r ampli/requirements.txt' to your ~/my/project/requirements.txt
 
-amplitude-analytics
+amplitude-analytics == 1.1.0
```
bohan-amplitude, line 26: Can we make the requirements version >= 1.1.0?

bohan-amplitude (approved):

PR #142, opened 2022-09-19
feat: update ingestion metadata js examples

liuyang1520:
## Notes
This change adds the ingestion metadata example for JS Ampli examples.

Using spread operation to assign the ingestion metadata value:
- doesn't get compile error in my local
- puts the value in events if using SDK supporting this field, otherwise the field is ignored

## Updates
- feat: update ingestion metadata js examples

## Tests
- Local echo server test


yuhao900914 (commented):
Nice!

qingzhuozhen (approved):
Seems JS/TS is easier, and pass in wouldn't affect the earlier versions?

liuyang1520:
> Seems JS/TS is easier, and pass in wouldn't affect the earlier versions?

It won't affect earlier versions, as adding unrecognized fields in config doesn't have any impact, tested in my local with both versions. Still investigating React-Native (legacy) one, will update Node and RN soon.

browser/javascript/v1/react-app/src/ampli/index.js
```diff
@@ -133,6 +133,12 @@ export const DefaultOptions = {
     branch: 'main',
     source: 'browser-js-ampli',
     versionId: '79154a50-f057-4db5-9755-775e4e9f05e6'
+  },
+  ...{
```
justin-fiedler, line 137: Why destructure here?

justin-fiedler (changes requested):
@liuyang1520 are these changes available in the backend codegen? We generally only update the examples via `ampli pull` once the codegen updates are already merged.

yuhao900914:
> @liuyang1520 are these changes available in the backend codegen? We generally only update the examples via `ampli pull` once the codegen updates are already merged.

I think ideally, all ampli changes will be in the same codegen pr. 
But for now we are trying to create a draft pr to align the method with the team.

browser/javascript/v1/react-app/src/ampli/index.js
yuhao900914: It won't have the type error using this way.

liuyang1520: Yep, this is to avoid the type error raised during the compilation. These changes are not in backend codegen yet. As @yuhao900914 mentioned, we are trying to create these PRs to present the final results codegen should generate.

justin-fiedler (approved):

PR #143, opened 2022-09-21
feat: update ingestion metadata jre examples

liuyang1520:
## Notes
This change adds the ingestionMetadata to event with the backwards compatibility support.

This example uses Java reflection, which is working for both old version and new version Java/Kotlin SDKs, including:
- jre-java v1
- jre-kotlin v1
- android-java v1
- android-java v2
- android-kotlin v1
- android-kotlin v2

## Updates
- feat: update ingestion metadata jre examples

## Tests
- echo server with both old/new version Java SDKs

### DIFF
--- a/android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
+++ b/android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
@@ -16,13 +16,14 @@
 
 package com.amplitude.ampli
 
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
 import com.amplitude.api.Amplitude
 import com.amplitude.api.AmplitudeClient
 import com.amplitude.api.MiddlewareExtra
 import com.amplitude.api.Plan
-import org.json.JSONArray
-import org.json.JSONException
-import org.json.JSONObject
 
 abstract class Event<E: Event<E>>(
     val eventType: String,
@@ -530,8 +531,7 @@ open class Ampli {
             val ingestionMetadata = clazz.newInstance()
             setSourceNameMethod.invoke(ingestionMetadata, "android-kotlin-ampli")
             setSourceVersionMethod.invoke(ingestionMetadata, "1.0.0")
-            val setIngestionMetadata =
-                AmplitudeClient::class.java.getMethod("setIngestionMetadata", clazz)
+            val setIngestionMetadata = AmplitudeClient::class.java.getMethod("setIngestionMetadata", clazz)
             setIngestionMetadata.invoke(client, ingestionMetadata)
         } catch (e: ClassNotFoundException) {
             println("com.amplitude.api.IngestionMetadata is available starting from Android SDK 2.38.2 version")
--- a/android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
+++ b/android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
@@ -16,14 +16,13 @@
 
 package com.amplitude.ampli
 
-import org.json.JSONArray
-import org.json.JSONException
-import org.json.JSONObject
-
 import com.amplitude.api.Amplitude
 import com.amplitude.api.AmplitudeClient
 import com.amplitude.api.MiddlewareExtra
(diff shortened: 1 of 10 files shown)


jre/kotlin/AmpliApp/build.gradle
```diff
@@ -11,7 +11,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.amplitude:java-sdk:1.8.0'
+    implementation 'com.amplitude:java-sdk:1.10.1'
```
yuhao900914, line 14: should we make it 1+?

liuyang1520: Thanks! Make sense, updated!

android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -519,6 +518,30 @@ open class Ampli {
         }
 
         this._client?.setPlan(options?.client?.plan ?: observePlan)
+
+        // set IngestionMetadata with backwards compatibility, min Android SDK version 2.38.2.
+        try {
+            val clazz = Class.forName("com.amplitude.api.IngestionMetadata")
```
justin-fiedler, line 524: @liuyang1520 @yuhao900914 @qingzhuozhen 
In general I would like us to make codegen as dumb and decoupled as possible. Anything we can move to the static SDKs (that can be unit tested, etc) is preferable to adding to codegen. I release we did similar type changes in python, js/ts but this implementation feels more tightly coupled and could lead to maintenance issues long term.

For these reasons I think the plugin/extra approach is preferable. Happy to discuss more.

justin-fiedler (changes requested):

android/kotlin/v1/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
liuyang1520: Thanks for reviewing! I reply here for @qingzhuozhen 's question too.

This solution includes the Kotlin Android SDK v1/v2 change. We tried the reflection on Kotlin Android v2 at first, and thought it doesn't work. After we figured out the reflection Java SDK, we found out the solution for Kotlin yesterday with the inspiration of Java SDK.

I prefer the reflection way, due to the following reasons:
- No other releases needed for SDKs, also legacy SDKs are using middleware, which need more efforts to explore where/how to attach the information
- Clean implementation, doesn't need to manipulate the `extra` field (I thought that was the only solution before), I think technically this field should be used by customer for adding information, this way we can avoid using it
- Low maintenance efforts, if anything goes wrong with ingestion metadata, this would be the single place to check in Ampli codegen. In the future, if we assure customers are using the latest version SDKs, we can safely replace this with the direct assigning of ingestion metadata, no need to update SDK anymore

The ingestion metadata functionality has been unit tested in static SDKs, this is to use it with supporting the backwards compatibility. In the long term, think we can explore a better way to release the codegen change (another story).

Sorry I haven't spent many efforts to investigate the reflection in the beginning, I think I should have tried that more before drawing the conclusion.

Happy to discuss/address more concerns! Thank you for reviewing!

android/kotlin/v2/AmpliApp/app/src/main/java/com/amplitude/ampli/Ampli.kt
```diff
@@ -41,6 +42,24 @@ class LoadClientOptions(
 
 val defaultObservePlan = Plan("main", "kotlin-ampli-v2", "0", "79154a50-f057-4db5-9755-775e4e9f05e6")
 
+class SetAmpliExtrasPlugin : Plugin {
```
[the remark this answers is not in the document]
liuyang1520: Thanks for reviewing! I don't think we have a EnrichmentPlugin class. The plugin is categorized by the `type` field. Let me know if I miss anything here.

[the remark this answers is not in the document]
yuhao900914: Just checked. I think you are right. Sorry for the confusion.

justin-fiedler (approved):

liuyang1520:
Close this, refer to the new one: https://github.com/amplitude/ampli-examples/pull/146

PR #147, opened 2022-10-12
AMP-61971 Ampli JS/TS v1 codegen: pass user_properties to underlying SDK

falconandy:


justin-fiedler (approved):
Looks great. Thanks @falconandy

PR #151, opened 2022-10-18
build: adds workflow to create jira issues

kevinpagtakhan:


yuhao900914 (approved):

.github/workflows/jira-issue-create.yml
```diff
@@ -0,0 +1,21 @@
...
+  workflow_call:
+    inputs:
+      label:
+        type: string
+
+jobs:
+  call-workflow-passing-data:
+    uses: amplitude/Amplitude-TypeScript/.github/workflows/jira-issue-create-template.yml@main
+    with:
+      label: "ampli"
```
liuyang1520, line 16: Minor, maybe "ampli examples"? I think we have another repo https://github.com/amplitude/ampli.

liuyang1520: nvm, think we are good

PR #157, opened 2022-11-17
DXOC-377 Add sessionId to EventOptions for jre ampli v1

bohan-amplitude:
PR for codegen https://github.com/amplitude/javascript/pull/21994
Tickets for customer request https://amplitude.atlassian.net/browse/DXOC-377

### DIFF
--- a/.github/workflows/ampli-merge-check.jre.yml
+++ b/.github/workflows/ampli-merge-check.jre.yml
@@ -8,9 +8,11 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


falconandy (changes requested):
1. Codegen also affects (partially, `EventOptions` only) Android v1. If it is expected, could you provide full support for Android (`createAmplitudeEvent`)? If not expected, could you remove partial Android support to not confuse users?
2. Not updated codegen test snapshots.

justin-fiedler (approved):

PR #158, opened 2022-11-19
AMP-62030 Call setGroup/groupIdentify from Amplitude SDKs

falconandy:
Updated examples to call `setGroup`/`groupIdentify` directly from Amplitude SDKs
ios examples (Swift/Objective-C) not tested.


.github/workflows/ci-build-react-native-test-v1.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [ 12.x ]
+        node-version: [ 14.x ]
```
justin-fiedler, line 14: Thank for updating this. Maybe we should go to lowest active e.g. 16?

android/java/v1/AmpliApp/app/src/test/java/com/example/ampliapp/AmpliTest.java
```diff
@@ -92,92 +91,6 @@ public void identify() {
         );
     }
 
-    @Test
```
justin-fiedler, line 95: Even though this is not part of the Ampli SDK, it could be good to keep the group tests, so that we know we can support the functionality via `ampli.client.<group-method>()`

justin-fiedler (changes requested):

android/java/v1/AmpliApp/app/src/test/java/com/example/ampliapp/AmpliTest.java
falconandy: Restored.

.github/workflows/ci-build-react-native-test-v1.yml
falconandy: Based on https://endoflife.date/nodejs and https://github.com/nodejs/release#release-schedule lowest `Active Support` version is 18,  lowest `Security Support` is 14.

justin-fiedler (approved):

PR #209, opened 2023-03-16
DXOC-652 set client.userId from Ampli options.userId

justin-fiedler:
https://amplitude.atlassian.net/browse/DXOC-652

Changes:
* Fixed Ampli iOS Tests, up til now they were completing before the Middleware (containing the checks) was run
* Added (temporary) fix for applying option.userId, will include in another PR


ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.m
```diff
@@ -674,7 +677,9 @@ - (void)identify:(NSString *_Nullable)userId event:(Identify *)event options:(Ev
     if (![self isInitializedAndEnabled]) {
         return;
     }
-    if (userId != nil) {
+    if (options != nil && options.userId != nil) {
```
falconandy, line 680: LGTM but does it help in our case?
I think, here
image
`b.userId` is a property added by the customer , not a property of `EventOptions`

I think, possible solution is to re-write customer's function
image
to
```
Ampli.instance().identify(userProperty.userId, event: userProperty)
```

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleAppTests/AmpliTests.m
```diff
@@ -20,14 +21,34 @@ - (void)setUp {
     [_ampli load:[LoadOptions builderBlock:^(LoadOptionsBuilder *b) {
         b.apiKey = @"test-api-key";
     }]];
+    _ampli.client.eventUploadThreshold = 1;
+    _hasMiddlewareRun = NO;
+    [_ampli flush];
+}
+
+- (void) waitForMiddlewareToRun {
```
falconandy, line 29: I would use deterministic `XCTestExpectation` to wait for middleware completion.
Additionally: new Amplitude instance for each test with new fresh storage (maybe fake/in-memory)
It looks like Swift unit tests in CI fail for similar reason - the example app is started with empty api key.

falconandy (changes requested):

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.m
justin-fiedler: Thanks Andrey, 100% agree but wanted to fix the EventOptions support before responding so that I could tell them they can set userId and deviceId via EventOptions and to use that instead of clouding up the Identify user_properties.

ios/objective-c/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleAppTests/AmpliTests.m
justin-fiedler: Thanks, I tried `XCTestExpectation` but it caused the tests to run multiple instances and threading issues. I am going to dig in more today.

justin-fiedler: Got it working with `XCTestExpectation`. Threading issues were resolved by using new instances of Ampli and Amplitude for each test.

PR #215, opened 2023-04-19
Amp 73622 update segment plugin example to support groups page

falconandy:


justin-fiedler (approved):
Looks great. Thanks @falconandy!

PR #218, opened 2023-05-14
DXOC-831 Actualized codegen

falconandy:
Codegen was outdated - updated.


jre/kotlin/AmpliApp/src/main/kotlin/com/amplitude/ampli/Ampli.kt
```diff
@@ -749,6 +719,42 @@ open class Ampli {
         this.track(userId, event, options, extra)
     }
 
+    /**
```
justin-fiedler, line 722: This file still shows "Identify" https://github.com/amplitude/ampli-examples/blob/3b643ec6f7d77c46ed66663a465c696f6f454c32/jre/kotlin/AmpliApp/src/main/kotlin/com/amplitude/ampli/Ampli.kt#L66

justin-fiedler (changes requested):
Thanks for updating all the projects. See comment about JRE Kotlin Identify.

jre/kotlin/AmpliApp/src/main/kotlin/com/amplitude/ampli/Ampli.kt
falconandy: This PR doesn't contain identify-related changes - it only updates outdated generated files to simplify next diffs in https://github.com/amplitude/ampli-examples/pull/219.

justin-fiedler (approved):

PR #282, opened 2023-10-02
AMP-85465 Objective-C v2 app

falconandy:
Example App for iOS Objective-C V2 runtime


ios/objective-c/v2/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.m
```diff
@@ -0,0 +1,811 @@
...
+ * [View Tracking Plan](https://data.amplitude.com/test-codegen/Test%20Codegen/events/main/latest)
+ *
+ * [Full Setup Instructions](https://data.amplitude.com/test-codegen/Test%20Codegen/implementation/obj-c-ampli-v2)
+ */
+
+#import <Foundation/Foundation.h>
+@import AmplitudeSwift;
+#import "Ampli.h"
+
+@implementation Event: NSObject
```
justin-fiedler, line 21: Can we make this extend `BaseEvent` rather than create a new class? Similar to Kotlin, TS, etc.

I would like to make it easy to use Ampli/Amplitude SDKs interchangably e.g. `amplitude.track(AmpliEvent(prop = true))`

Same for Swift codegen.

justin-fiedler (changes requested):
Thanks @falconandy. Overall looks good. See comment about extending BaseEvent.

ios/objective-c/v2/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/Ampli/Ampli.m
falconandy: Ok, I'll try. For Swift it can introduce some breaking changes.

justin-fiedler: Thanks @falconandy. We are still in beta, so it's ok if you need breaking changes in Swift.

falconandy: Swift: done.
Objective-C: Subclassing Swift classes (`AMPBaseEvent`) from Objective-C is not supported (https://forums.swift.org/t/cannot-subclass-a-class-with-objc-subclassing-restricted-attribute/5674)

justin-fiedler: Thanks @falconandy for ObjC could we write [ObjCBaseEvent.swift](https://github.com/amplitude/Amplitude-Swift/blob/main/Sources/Amplitude/ObjC/ObjCBaseEvent.swift) in ObjC to work around this issue?

falconandy: @justin-fiedler I changed codegen - https://github.com/amplitude/ampli-examples/commit/77d9e041cc7cc41a06daf0c6151e98217e969caf
WDYT?

justin-fiedler: Thanks @falconandy.

1. I think this works ok but would make setting schema based properties hard after initial creation via the static builder.

```
AMPBaseEvent *event = [EventWithAllProperties requiredBool:true ....]
[event.eventProperties.set:@"requiredBool" value: false]
```

2. Another option could be to keep the previous Ampli `Event` class and add a new `asBaseEvent(): AMPBaseEvent` but that also seems clunky and would like to avoid the extra type in the generated code.

3. Since we are declaring the new class ObjCBaseEvent class specific for ObjC usage I think that we could port ObjCBaseEvent implementation to be in Obj-C and use `NS_SWIFT` and extensions (if needed). See Itly SDK example here
* https://github.com/amplitude/itly-sdk-ios/tree/master/Sdk/Sdk

Could you please try to see if option 3 could work? Thanks

falconandy: @justin-fiedler About option 1:
Could you please clarify more - I think, nothing was changed here in comparison with Objective-C V1?

justin-fiedler: @falconandy good point. I was thinking of TS which allows for easy setting of the generated event properties. I agree this provides the same functionality as Kotlin, Swift.

I think in this case the current implementation (option 1) is good. Will review again, but this this solution should work. Thanks

ios/objective-c/v2/AmpliObjectiveCSampleApp/AmpliObjectiveCSampleApp/AppDelegate.m
```diff
@@ -0,0 +1,167 @@
...
+    [identifyArgs set:@"optionalString" value:@"optional string"];
+    [ampli.client groupIdentify:@"ampli group type" groupName:@"ampli objective-c group" identify:identifyArgs];
+
+    // Track events with dedicated event methods
+    [ampli track:[EventNoProperties build]];
+
+    [ampli track:[EventMaxIntForTest intMax10: 5]];
+    [ampli track:[EventMaxIntForTest intMax10: 20]];
+
+    [ampli track:[EventWithConstTypes build]];
```
justin-fiedler, line 90: nit: can we rename `build` `new`? I feel that might be a little more idiomatic. wdyt?

falconandy: we can but XCode/compiler shows a warning - `new` is expected to return its class type
image

justin-fiedler: Thanks for checking. `build` is good.

justin-fiedler (approved):
Thanks @falconandy this is great!

